PR #29, opened 2017-05-02
Fix the weird scrollbar jumping issue

Copons:
The issue was caused by the note list not having a height value on component mount, likely due to the notification panel being hidden.

This introduce a new forced height calculation to be performed whenever the panel is about to open.

### DIFF
--- a/src/Notifications.jsx
+++ b/src/Notifications.jsx
@@ -1,14 +1,14 @@
-import React, { PropTypes, PureComponent } from 'react';
-import { Provider } from 'react-redux';
+import React, {PropTypes, PureComponent} from 'react';
+import {Provider} from 'react-redux';
 
-import { store } from './state';
+import {store} from './state';
 import actions from './state/actions';
 
 import RestClient from './rest-client';
-import { setGlobalData } from './flux/app-actions';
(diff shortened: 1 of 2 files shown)


src/Notifications.jsx
```diff
@@ -82,6 +82,7 @@ export class Notifications extends PureComponent {
                         global: globalData,
                         client,
                         locale: this.props.locale,
+                        isShowing: this.props.isShowing,
```
dmsnell: alphabetical order! :smile:

Copons: G → C → L → I

It does make sense to me!

src/templates/index.jsx
```diff
@@ -151,6 +151,11 @@ const Layout = React.createClass({
             this.noteListTop = noteList.scrollTop;
         }
 
+        // set the list height if it's not set yet
```
dmsnell: could we explain why this could be the case? it seems obvious from the code _that_ this is setting it but what I would care about is _why didn't it set it?_

something like…
> if the panel is closed when the component mounts then the calculated height will be zero because it's hidden. we need to make sure that when the panel opens, if we found that zero value, that we appropriately set the real height when actually rendered.

PR #34, opened 2017-05-03
Notification: Remove scrollTo occurence in comment reply

vindl:
Judging by the comments we can remove this since it was specific to `iFrame` and iPhone.

### Testing instructions 

TBD

### DIFF
--- a/src/templates/comment-reply-input.jsx
+++ b/src/templates/comment-reply-input.jsx
@@ -145,12 +145,6 @@ const CommentReplyInput = React.createClass({
                 rowCount: 1,
             });
         }
-
-        // Force scroll the parent window to the top on iPhone
-        // Note: This can be removed if we move away from using an iFrame for the notifications window
-        if (window.parent && hasTouch()) {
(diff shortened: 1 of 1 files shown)


src/templates/comment-reply-input.jsx
```diff
@@ -145,12 +145,6 @@ const CommentReplyInput = React.createClass({
                 rowCount: 1,
             });
         }
-
-        // Force scroll the parent window to the top on iPhone
-        // Note: This can be removed if we move away from using an iFrame for the notifications window
-        if (window.parent && hasTouch()) {
```
vindl: @dmsnell what do you think about adding a check here to see if this is loaded in an `iFrame`, and conditionally perform the `scrollTo` based on that?

dmsnell: I'd be fine with that. we used to check for the presence of the iframe in other places. ideally if we can accomplish with without knowing I'd prefer that but whatevs.

what is the challenge in not acknowledging the iframe?

PR #35, opened 2017-05-03
Remove unnecessary scroll to top of the window on close panel

Copons:
`scrollToTop()` is not the intended behaviour.

The idea was that, when closed, the panel scrollbar was being taken back to the top.
Now instead the scrollbar stays at the same position after closing the panel.

### DIFF
--- a/src/state/action-middleware/ui/index.js
+++ b/src/state/action-middleware/ui/index.js
@@ -8,8 +8,6 @@ import { onTogglePanel } from '../public-api';
 
 import { findNextNoteId } from '../../../templates';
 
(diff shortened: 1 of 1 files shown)


dmsnell:
Let's check for this behavior:

 1. Select long note in narrow detail view
 2. Scroll note in detail view
 3. Go back to list view
 4. Select another long note in narrow detail view
 5. Check that note renders at top scroll, not already scrolled
 6. However, if note is same as before, leaving pre-scroll is okay

Copons:
@dmsnell 
> Let's check for this behavior

Taken care in https://github.com/Automattic/notifications-panel/pull/38

PR #37, opened 2017-05-03
Replace following noticon with Gridicon

drw158:


drw158:
Yea, I don't disagree. I think the following icon should be green.

PR #38, opened 2017-05-03
Detail view scrolls to top when open

Copons:
Whenever a note detail is opened, scroll it back to top, unless it's the same as the previously opened note (if any).

Also sorry for the prettier changes.

### DIFF
--- a/src/templates/index.jsx
+++ b/src/templates/index.jsx
@@ -90,7 +90,7 @@ const Layout = React.createClass({
         return {
             lastSelectedIndex: 0,
             navigationEnabled: true,
-            previouslySelectedNote: null,
+            previouslySelectedNoteId: null,
             selectedNote: null,
         };
     },
@@ -128,7 +128,7 @@ const Layout = React.createClass({
     componentWillReceiveProps: function(nextProps) {
         if (this.props.selectedNoteId) {
             this.setState({
-                previouslySelectedNote: this.props.selectedNoteId,
+                previouslySelectedNoteId: this.props.selectedNoteId,
             });
         }
 
@@ -183,6 +183,13 @@ const Layout = React.createClass({
 
         if (!hasNote) {
             this.props.unselectNote();
+        }
+    },
+
+    componentDidUpdate: function() {
+        const { previouslySelectedNoteId, selectedNote } = this.state;
+        if (this.detailView && selectedNote !== previouslySelectedNoteId) {
+            this.detailView.scrollTop = 0;
         }
     },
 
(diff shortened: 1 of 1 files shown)


src/templates/index.jsx
```diff
@@ -125,6 +126,12 @@ const Layout = React.createClass({
     },
 
     componentWillReceiveProps: function(nextProps) {
+        if (this.props.selectedNoteId) {
+            this.setState({
+                previouslySelectedNote: this.props.selectedNoteId,
```
dmsnell: can we keep the `id` terminology the same so that we don't confuse objects with ids?

we have made this mistake in this app before and it's unfortunately caused much grief. that is to say that we have had situations where we had a valid note object but it wasn't in the list so the id was useless. I would much prefer here to track just the id and indicate via the name that such is the case: `previouslySelectedNoteId` or something

src/templates/index.jsx
```diff
@@ -453,13 +461,23 @@ const Layout = React.createClass({
...
     render() {
+        const { previouslySelectedNote, selectedNote } = this.state;
+
         const currentNote = find(
             this.props.notes,
             matchesProperty('id', this.props.selectedNoteId)
         );
         const filteredNotes = this.filterController.getFilteredNotes(this.props.notes);
 
+        if (this.detailView && selectedNote !== previouslySelectedNote) {
```
dmsnell: parens here would be helpful. it's not obvious to me that JS will do what we want

```js
(this.detailView && selectedNote) !== previouslySelectedNote

// vs.

this.detailView && ( selectedNote !== previouslySelectedNote )
```

src/templates/index.jsx
```diff
@@ -453,13 +461,23 @@ const Layout = React.createClass({
         this.noteListVisibilityUpdater = updater;
     },
 
+    storeDetailViewRef(ref) {
+        this.detailView = ref;
+    },
+
     render() {
+        const { previouslySelectedNote, selectedNote } = this.state;
```
dmsnell: is the currently selected note not coming through via props?

src/templates/index.jsx
```diff
@@ -453,13 +461,23 @@ const Layout = React.createClass({
...
+
         const currentNote = find(
             this.props.notes,
             matchesProperty('id', this.props.selectedNoteId)
         );
         const filteredNotes = this.filterController.getFilteredNotes(this.props.notes);
 
+        if (this.detailView && selectedNote !== previouslySelectedNote) {
+            this.detailView.scrollTop = 0;
```
dmsnell: hm. is it common to adjust scroll in render? we might re-render but not want to scroll for some reason, such as when a new note appears from the server. would this be more appropriate somewhere like `componentWillUpdate` or `componentDidUpdate`?

src/templates/index.jsx
Copons: > Just when I thought I added them in, prettier took them back out!

PR #40, opened 2017-05-03
Add Sticky Headers

roundhill:
Removes the js based sticky headers and replaces it with `position: sticky`. Unfortunately, Chrome doesn't seem to support sticky headers with a fixed position parent, so we've added a browser check for it in order to disable the stickies.


roundhill:
> Cry for Chrome
😢 😿

PR #44, opened 2017-05-03
Fix detail view scroll position after HTML changes

Copons:
Previously the detail view scrollbar was on its root element.
It was moved to the `<ol>` rendering #38 futile.

While the main element has a `max-height: 100%`, making keeping the scroll position an automatic task, the constantly changing list height introduces the need to keeping track of the scroll position when closing the detail view.

This PR moves the ref to the correct element and stores the detail view list scroll position.
The scroll position will be used again if the same note is opened again immediately.

### DIFF
--- a/src/templates/index.jsx
+++ b/src/templates/index.jsx
@@ -90,8 +90,8 @@ const Layout = React.createClass({
         return {
             lastSelectedIndex: 0,
             navigationEnabled: true,
+            previousDetailScrollTop: 0,
             previouslySelectedNoteId: null,
-            previouslySelectedNoteScrollTop: 0,
             selectedNote: null,
         };
     },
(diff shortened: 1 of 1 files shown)


src/templates/index.jsx
```diff
@@ -187,9 +189,18 @@ const Layout = React.createClass({
...
+        }
+        const {
+            previouslySelectedNoteId,
+            previouslySelectedNoteScrollTop,
+            selectedNote,
+        } = this.state;
+        if (selectedNote !== previouslySelectedNoteId) {
             this.detailView.scrollTop = 0;
+        } else {
+            this.detailView.scrollTop = previouslySelectedNoteScrollTop;
```
dmsnell: how about we only set the scroll top once and use a ternary to assign the new value?

```js
this.detailView.scrollTop = selectedNote !== previouslySelectedNoteId
	? previouslySelectedNoteScrollTop
	: 0;
```

PR #47, opened 2017-05-03
Replace time noticon with gridicon

drw158:
This replaces the clock icon.

screen shot 2017-05-04 at 2 57 01 pm

I also cleaned up the css a bit by using the existing %headertext extend.


src/templates/gridicons.jsx
```diff
@@ -223,6 +223,13 @@ export default React.createClass({
                     </svg>
                 );
 
+                case 'gridicons-time':
```
roundhill: Spacing: This looks like one tab too many here?

PR #52, opened 2017-05-03
Build: provide both a standalone js bundle and build bundle

gwwar:
This PR updates package.json to explicitly list `peerDependencies` and published files. Here we also update webpack, so we can have a final standalone js file, and one for just the component.

Before I debug further, I wanted to check in on this approach for packaging. Calypso runs, but we run into the following js errors with the transpiled file:
screen shot 2017-05-03 at 4 12 23 pm

### Testing Instructions
- Checkout this branch
- `npm install`
- `make build`
- `npm link`
- Navigate to wp-calypso
- Checkout `update/load-notes-client-as-npm-module` from https://github.com/Automattic/wp-calypso/pull/13583
- In the wp-calypso folder call `npm link notifications-panel`
- In the wp-calypso call `make run`


gwwar:
Closing this out since I probably need a bit of time to get this going properly

PR #60, opened 2017-05-04
Notifications: fix comment reply suggestions scroll

vindl:
`ensureSelectedSuggestionVisibility` was not working properly, and was using `window.scrollTo`.
Fixed the condition to check visibility and uses current element to scroll into view instead.


vindl:
Note: this doesn't work as expected in production since this condition is wrong:
```javascript
 if (offsetTop - window.pageYOffset <= 0)
```
It's true when suggestion is already in view, so there is no need to scroll in that case.

PR #66, opened 2017-05-04
Fix visual glitches in Calypso

### LINKED ISSUE #56: Visual glitch in stats
It seems like some CSS leak is causing broken layout in Stats visitors: 

warnings

This is displayed fine in Calypso master.

Copons:
Fixes #56 

Namespace the form extend classes to prevent CSS leaking to Calypso.


src/boot/stylesheets/shared/forms.scss
```diff
@@ -20,72 +9,40 @@
...
-fieldset,
-input[type="text"],
-input[type="email"],
-input[type="number"],
-input[type="password"],
-textarea,
-select,
-label {
-  box-sizing: border-box;
-}
```
vindl: Missing newline.

src/boot/stylesheets/shared/forms.scss
```diff
@@ -20,72 +9,40 @@
...
   }
 }
 
-%textarea {
+%wpnc-textarea {
   min-height: 92px;
 }
-
-
-// ==========================================================================
```
vindl: Will the removal of this block affect `iFrame` notes versions?

Copons: Tested it and it works, but I'd be much more comfortable with a second look on that!

vindl: Maybe if we just keep everything in this file as is, but nest it in `wpnc_main`?

PR #91, opened 2017-05-09
Fix/notification bell

gwwar:
This PR attempts to resolve #13772 where the notifications bell can get into a stuck state, even after we've read all notifications.

This is part of https://github.com/Automattic/wp-calypso/pull/13784 to test, use instructions in 13784

PR #92, opened 2017-05-09
New notification announcement

rodrigoi:
When receiving a new message from pinghub, the bell icon does not update it's status, caused by `hasSeenNotes` initial value of `false`.

Before:
before

After:
after

@dmsnell this will rollback https://github.com/Automattic/notifications-panel/pull/90

To test:
run `localStorage.setItem( 'debug', 'notifications:rest-client' );` to enable debug messages and send a new notification.

### DIFF
--- a/src/rest-client/index.js
+++ b/src/rest-client/index.js
@@ -299,8 +299,6 @@ function getNotesList() {
     });
 }
 
-let hasSeenNotes = false;
-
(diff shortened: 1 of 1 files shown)


rodrigoi:
Closed in favor of https://github.com/Automattic/notifications-panel/pull/91

PR #94, opened 2017-05-09
Notes: fix list when notification comes in before panel is opened

gwwar:
This fixes https://github.com/Automattic/wp-calypso/issues/13835 where notifications would not fully load if a new notification was received before the panel was opened.

spinner

### Testing Instructions
Use https://github.com/Automattic/wp-calypso/pull/13843 to test behavior

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.1.7",
+  "version": "1.1.8",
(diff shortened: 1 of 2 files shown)


roundhill:
Tested and confirmed that it fixes the issue :shipit:

gwwar:
Thanks for the reviews @roundhill @rodrigoi

PR #95, opened 2017-05-09
Fixing a one pixel shift when opening a notification in the filter bar.

roundhill:
I noticed that the filter bar was shifting over 1 pixel when we removed the border from the left side. This adds a 1px white border when the detail view is showing which fixes the shift.

@Copons can you please review? I also wrapped the border you added previously in a media query so that it doesn't show when the list view is smaller than 800px.

Before:
before

After:
after

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -181,12 +181,12 @@ body {
     		right: 0px;
     	}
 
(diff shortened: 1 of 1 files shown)


roundhill:
> Though I'd change both media queries to 661px, since the panel goes width: 100% at 660px.

Ah right, I had it in my head that it broke at 800 for some reason. Updated!

Issue #103, opened 2017-05-15
Unread items disappear once you click on them while the unread filter is selected


drw158:
> The big one is that “unread” is basically broken, when you click things they disappear, which loses the relationship between the item and the detail view, and also breaks keyboard navigation

Screenshot of problem:

![screen shot 2017-05-15 at 11 02 34 am](https://cloud.githubusercontent.com/assets/618551/26066906/0f727240-395e-11e7-8eea-12faeb62e752.png)

To reproduce:

- receive a notification
- click on Unread filter
- click on an unread item
- you'll see something similar to the above, where the item is no longer there, but the detail view is still present

Before, I think we had some sort of mechanism for keeping the item visible until the user changed filters.

drw158:
I can confirm that the keyboard navigation is sort of broken. To reproduce:

- receive a notification
- click on Unread filter
- use your down arrow key to select the first unread note
- notice that the blue border is visible
- the detail view does not change
- notice that once you click on the note with your mouse, the detail view updates correctly 

![screen shot 2017-05-15 at 11 14 12 am](https://cloud.githubusercontent.com/assets/618551/26067378/a71c4e80-395f-11e7-8278-e4e52c27b843.png)

roundhill:
I'll try and tackle this one.

drw158:
Can confirm this is a regression. Check out the `notifications` widget url and you can see that it works there, but not in `notificationsbeta`.

roundhill:
Fixed via #121

PR #105, opened 2017-05-15
Keep a transient list of unread notes so they don't disappear from the unread filter

roundhill:
WIP: First attempt at keeping the notes in the unread filter from disappearing. #103

@dmsnell I've Kinda thrown this together, wondering if you have any other ideas for this? 

~~~There's probably some use cases that this would cause a problem with. Essentially you'll always get the same notes list until you switch away from the unread filter...~~~
Updated to use the `notesRead` list and added the filter bar to the state.

Here it is in action:
![unread-comments-stay](https://cloud.githubusercontent.com/assets/789137/26070839/e7f16ada-395a-11e7-92e0-a063a7bd2a67.gif)


src/state/notes/reducer.js
```diff
@@ -51,8 +51,18 @@ export const noteLikes = (state = {}, { type, noteId, isLiked }) => {
...
+    if (types.SELECT_NOTE === type) {
+        return noteId ? { ...state, [noteId]: filterName } : state;
+    }
+
+    if (types.RESET_READS === type) {
+        return mapValues(state, () => '');
+    }
+
+    return state;
+};
```
roundhill: Hmm ok. As far as I could tell this state was only keeping track of notes that were marked read locally, and was always set to `true`.

I believe in the previous version of the client we kept a separate array of notes that were read when on the unread filter, should we just do that again?

src/templates/index.jsx
```diff
@@ -299,7 +300,7 @@ const Layout = React.createClass({
 
         /* If we are in detail view, move to next note */
         if (this.props.selectedNoteId) {
-            return this.props.selectNote(filteredNotes[newIndex].id);
+            return this.props.selectNote(filteredNotes[newIndex].id, this.props.filterName);
```
roundhill: Could just be my misunderstanding of redux :) Should I just access the filterName directly in the  reducer?

src/templates/filter-bar-controller.js
```diff
@@ -33,7 +39,20 @@ FilterBarController.prototype.getFilteredNotes = function(notes) {
...
+
+    // Prevent notes in the unread filter from disappearing when marked as read.
+    if (this.selected && this.selected.name === 'unread') {
+        const noteReads = store.getState().notes.noteReads;
+
+        const unreadsFilter = note => (note.id in noteReads && noteReads[note.id] === 'unread');
+        filteredNotes = notes.filter(unreadsFilter)
+            .concat(filteredNotes)
+            .sort((note1, note2) => (new Date(note2.timestamp) - new Date(note1.timestamp))
+        );
```
roundhill: Ahh nice, you're right. I updated it to a better version in 7a3ac5f. Although I think you might have some ideas on how to do that with fewer lines of code ;)

src/templates/index.jsx
dmsnell: > Should I just access the filterName directly in the reducer?

this is where middleware comes into play. sometimes we can handle it by proper restructuring of the data in the reducers.

the concept I was shooting for was that we can track a list of viewed notifications and flush that when we change filters. that list is the override of `unread`

```js
const recentlyRead = ( state = {}, { type, filterName, noteId } ) => {
	if ( SELECT_FILTER === type ) {
		return {};
	}

	if ( SELECT_NOTES === type ) {
		return { ...state, noteId: true };
	}

	return state;
}

const getRecentlyRead = state => Object.keys( state.recentlyRead )
```

^^^ something like that

this illustrates how Redux can lead us to think about…

> what changes this data

instead of

> what data does this change

the "recentlyRead" reducer is completely in control of its own state and watches for whichever global actions affect it. hope this helps!

dmsnell: > Should I just access the filterName directly in the reducer?

this is where middleware comes into play. sometimes we can handle it by proper restructuring of the data in the reducers.

the concept I was shooting for was that we can track a list of viewed notifications and flush that when we change filters. that list is the override of `unread`

```js
const recentlyRead = ( state = {}, { type, filterName, noteId } ) => {
	if ( SELECT_FILTER === type ) {
		return {};
	}

	if ( SELECT_NOTES === type ) {
		return { ...state, noteId: true };
	}

	return state;
}

const getRecentlyRead = state => Object.keys( state.recentlyRead )
```

^^^ something like that

this illustrates how Redux can lead us to think about…

> what changes this data

instead of

> what data does this change

the "recentlyRead" reducer is completely in control of its own state and watches for whichever global actions affect it. hope this helps!

dmsnell:
@roundhill if you consider what I wrote in the comment about the `recentlyRead` reducer we could update the `getNoteIsRead()` selector for the notes to simply take this one additional consideration. that would apply globally across the app when determining if a note is considered "read"

roundhill:
> @roundhill if you consider what I wrote in the comment about the recentlyRead reducer we could update the getNoteIsRead() selector for the notes to simply take this one additional consideration. that would apply globally across the app when determining if a note is considered "read"

Thanks, I should have thought about doing it that way in the first place. Updated in 2ed74b7. I'm still adding the filter name to the `noteReads` object in order to track notes that were read while in the unread filter.

roundhill:
I'm going to abandon this PR. After a discussion on Slack with @dmsnell I'm going to create two new PRs:

1. Converting the filter bar selection to the state.
2. Add an additional reducer that tracks the unread filter reads, instead of modifying the `noteReads` object like I did here

PR #107, opened 2017-05-17
Close detail view when changing filters

### LINKED ISSUE #101: Detail view open after changing filters
> if you have a detail view open, then switch to one of the filters, the sidebar stays open when it used to close, so you get where the left detail view doesn’t match what’s shown in the right.

roundhill:
Dispatch the `unselectNote` action when changing filters, so that the detail view closes in desktop mode.

**To Test**
* In the wide (two-panel) view, select a note. The detail view will open.
* Change to a different filter at the top of the list view (Unread, Comments, etc)
* The detail view should close.
* Also test small screen/mobile interaction works as expected.

Fixes #101

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.1.10",
+  "version": "1.1.11",
   "description": "The core notifications panel for WordPress.com notifications",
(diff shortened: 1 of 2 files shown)


roundhill:
Thanks for the review @gwwar.

Issue #104, opened 2017-05-15
Add a way to go to your notification settings from the panel


drw158:
There could be an easier, more direct way to access the notification settings.

One idea from Dan:

> Maybe a settings icon could go in the first time group header on the right? Lots of space there

roundhill:
Doesn't look too bad :)

<img width="410" alt="screen shot 2017-05-17 at 4 16 36 pm" src="https://cloud.githubusercontent.com/assets/789137/26180106/40b25324-3b1c-11e7-9bec-5223d02c29ad.png">

dmsnell:
Resolved by #108

PR #108, opened 2017-05-18
Add a settings button to the first time group header.

roundhill:
Opens the notifications settings when clicked which currently just opens them in a new window since that's about all we can do from the iframe version.

@dmsnell maybe you could add your middleware solution to get it working to open the route directly when in Calypso?

screen shot 2017-05-18 at 4 48 16 pm

Refs #104


dmsnell:
@roundhill found a bug where if there were no notes for `Today` then the settings icon wouldn't show and then I moved the window opening to middleware.

we're not yet at the coupling point with Calypso so I'm not sure if we can reasonably navigate Calypso to the settings from this PR as that will take some additional work to make happen.

dmsnell:
On second thought @roundhill I went ahead and added in the custom middleware idea. We can refine it but I think it will work. We will have to coordinate the version bump in Calypso with this because at the moment the settings icon won't do anything in Calypso until we create the corresponding custom middleware there.

src/state/index.js
```diff
@@ -13,6 +13,10 @@ const reducer = combineReducers({
...
 const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
-const withMiddleware = composeEnhancers(applyMiddleware(actionMiddleware))(createStore);
+const withMiddleware = customMiddleware =>
+    composeEnhancers(applyMiddleware(actionMiddleware(customMiddleware)))(createStore);
 
-export const store = withMiddleware(reducer, reducer(undefined, { type: '@@INIT' }));
+export let store = { dispatch: () => {}, getState: () => {} };
+
+export const init = ({ customMiddleware }) =>
+    (store = withMiddleware(customMiddleware)(reducer, reducer(undefined, { type: '@@INIT' })));
```
dmsnell: @gwwar @coderkevin see above for the most primitive version of my custom integration middleware injection idea

src/templates/gridicons.jsx
```diff
@@ -238,6 +238,20 @@ export default React.createClass({
                         </g>
                     </svg>
                 );
+
+            case 'gridicons-cog':
+                return (
+                    <svg {...sharedProps} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+                        <title>Arrow Left</title>
```
drw158: Should be `Cog`

dmsnell: Should be fixed

roundhill:
Tried it out in Calypso 😍

PR #109, opened 2017-05-22
Modify text rendering inside notes to add styling

dmsnell:
<del>Resolves automattic/wp-calypso#2044</del> (resolved via API)

<del>A long-standing issue has been the lack of spacing between paragraphs of
text. Every line-break has received a `<br />` tag to separate them but
this left paragraphs flowing without break.</del>

This patch is a hack that adds on some primitive styling in order to get
"the best bang for the buck" until a proper refactor of the text
processing occurs.

Namely:
 - Detect lists and make them real lists with `<ul>` and `<li>`
 - Detect emphasized headers like `Next Week: do something`
 - Split paragraphs of text vertically

The new `toBlocks` reducer does the real work here and takes a fairly
naive approach. Certain bugs exist but I believe that the work in this
patch will relieve much more frustration than it possibly introduces.

@drw158 can you review the use of the bullet? I know there are some styling things I just blasted in without knowing how to do them. also, I seem to remember that we may have intentionally prevented the use of `<li>` so if the bullet isn't right we can change that.

**Testing**
Boot up with `npm start` and do a visual inspection of the notes. Look for inconsistencies in the rendering or blatant problems.

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -655,14 +655,53 @@ body {
           display: block;
         }
       }
+
+      pre {
+        background: $gray-light;
+        border: 1px solid lighten($gray, 20%);
+        border-radius: 3px;
+        padding: 4px;
+
+        code {
+          border: none;
+          background: none;
+        }
+      }
+
+      code {
+        font-family: $code;
+        font-size: 90%;
+        color: darken($gray, 20%);
+        background: $gray-light;
+        border: 1px solid lighten($gray, 20%);
+        border-radius: 3px;
+        padding: 0 2px;
+      }
     }
 
     .wpnc__paragraph:first-of-type {
       margin-top: 0;
     }
 
+    .wpnc__body-list {
+      list-style: disc;
+      padding: 0 16px;
+    }
+
+    .wpnc__body-todo {
+      list-style: none;
+
+      .wpnc__todo-done:before {
+        content: '◉ ';
+      }
+
+      .wpnc__todo-not-done:before {
+        content: '◎ ';
+      }
+    }
+
     blockquote {
-      margin: 0 $padding-medium;
+      margin: 0 $padding-medium $padding-medium;
       font-style: italic;
       color: $gray;
       background: transparent;
--- a/src/templates/functions.jsx
+++ b/src/templates/functions.jsx
@@ -2,16 +2,138 @@ import React from 'react';
 
 import { html as toHtml } from '../indices-to-html';
 
+/**
+ * Adds markup to some common text patterns
+ *
+ *  - Bullet lists
+ *  - Todo lists, WP style
+ *  - Inline `code` snippets
+ *  - Code fences
+ *  - Header: description/explanation paragraphs
+ *
+ * Note: This code is only meant to serve until a
+ * proper parser can be built up to convert the
+ * unstructured text into structured data. Since at
+ * this time we still create HTML strings directly
+ * and on every render this function will serve
+ * sufficiently but it should not be looked upon
+ * as good example code!
+ *
+ * @param {String} text input list of blocks as HTML string
+ * @returns {String} marked-up text
+ */
+const toBlocks = text =>
+  text.split('\n').reduce(({ out, inFence, inList }, raw) => {
+    // detect code fences
+    // ```js
+    // doFoo()
+    // ```
+
+    // code fence?
+    // WordPress can replace `` with a fancy double-quote
+    if (/^(```|“`)[a-z1-9]*\s*$/i.test(raw)) {
+      // opening a fence
+      if (!inFence) {
+        return {
+          out: out + '<pre><code>',
+          inFence: true,
+          inList,
+        };
+      }
+
+      // closing a fence
+      return {
+        out: out + '</code></pre>',
+        inFence: false,
+        inList,
+      };
+    }
+
+    // content inside a fence
+    if (inFence) {
+      return {
+        out: out + raw + '\n',
+        inFence,
+        inList,
+      };
+    }
+
+    // emphasized definition-like text
+    // Some value: some description
+    // Header: Value
+    //
+    // Not! This is fun. This: again; isn't emphasized.
+    // May detect false positive if colon found in first sentence.
+    const defined = /^[\w\s-_]+:/.test(raw)
+      ? `<strong>${raw.split(':')[0]}:</strong>${raw.replace(/^[^:]+:/, '')}`
+      : raw;
+
+    // inline `code` snippets
+    const coded = defined.replace(/`([^`]+)`/, '<code>$1</code>');
+
+    // detect list items
+    //  - one
+    //  * two
+    //  [bullet] three
+    //  [dash] four
+    if (/^\s*[*\-\u2012-\u2015\u2022]\s/.test(coded)) {
+      return {
+        out:
+          out +
+            (inList ? '' : '<ul class="wpnc__body-list">') +
+            `<li>${coded.replace(/^\s*[*\-\u2012-\u2015\u2022]\s/, '')}</li>`,
+        inFence,
+        inList: true,
+      };
+    }
+
+    // detect todo lists
+    //  x Done
+    //  o Not done
+    //  O Also not done
+    // X also done
+
+    if (/^\s*x\s/i.test(coded)) {
+      return {
+        out:
+          out +
+            (inList ? '' : '<ul class="wpnc__body-todo">') +
+            `<li class="wpnc__todo-done">${coded.replace(/^\s*x\s/i, '')}</li>`,
+        inFence,
+        inList: true,
+      };
+    }
+
+    if (/^\s*o\s/i.test(coded)) {
+      return {
+        out:
+          out +
+            (inList ? '' : '<ul class="wpnc__body-todo">') +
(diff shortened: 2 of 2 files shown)


roundhill:
Fixed on the api side by @dmsnell 👍

dmsnell:
@roundhill would still like some feedback here for bullet lists and emphasized "header" text

think it's worth using?

roundhill:
> @roundhill would still like some feedback here for bullet lists and emphasized "header" text
> 
> think it's worth using?

Oh! Sorry for prematurely closing this. I think it might be worth adding these additional formatting niceties. Any thoughts @drw158?

src/boot/stylesheets/main.scss
```diff
@@ -690,12 +690,21 @@ body {
...
+			}
     	}
 
     	.wpnc__paragraph:first-of-type {
     		margin-top: 0;
     	}
 
+		.wpnc__body-list {
+			list-style: disc;
+			margin-left: 2em;
```
drw158: I would do this instead of margin-left

`padding: 0 16px;`

That'll line things up on the sides.

screen shot 2017-05-23 at 10 48 13 am

drw158:
The meeting note notification looks great:

screen shot 2017-05-23 at 10 48 13 am

On an older comment, for some reason the list tags are not used, and the bullet is being inserted directly.

dmsnell:
@drw158 thanks for the CSS suggestions. I added a new chunk for rendering inline code and I took direct inspiration from how Slack does it. you might want to review that.

@roundhill removed the paragraph detection and added todo lists as well!

ready for another review!

src/templates/functions.jsx
```diff
@@ -2,16 +2,96 @@ import React from 'react';
...
+ * unstructured text into structured data. Since at
+ * this time we still create HTML strings directly
+ * and on every render this function will serve
+ * sufficiently but it should not be looked upon
+ * as good example code!
+ *
+ * @param {String} text input list of blocks
+ * @returns {String} marked-up text
+ */
+const toBlocks = text =>
```
gwwar: Is text already stripped of html by this point? 

Overall this looks ok so far for a quick and dirty replace. I'd recommend adding unit tests for these cases so refactoring is pretty easy.

dmsnell: actually the comment is misleading, it's an HTML string so markup is already inside of it because of other processing in the app. the note data comes from the API without any HTML though.

src/boot/stylesheets/main.scss
```diff
@@ -690,12 +690,41 @@ body {
...
+				content: '✓ '
+			}
+
+			.wpnc__todo-not-done:before {
+				content: '✗ '
+			}
+		}
+
     	blockquote {
     		margin: 0 $padding-medium;
```
drw158: Right now blockquotes look weird because they do not have a bottom margin. I think before the last commit, they were wrapped in a `p` maybe.

Change margin to this and it should look better:

`margin: 0 $padding-medium $padding-medium;`

drw158:
Mentions seem to be broken, cutting off a lot of text:

screen shot 2017-05-24 at 10 50 46 am
screen shot 2017-05-24 at 10 50 20 am

src/boot/stylesheets/main.scss
```diff
@@ -690,12 +690,41 @@ body {
     				display: block;
     			}
     		}
+
+			code {
+				font: $code;
```
drw158: change to `font-family` to solve the invalid property error.

src/boot/stylesheets/main.scss
```diff
@@ -690,12 +690,41 @@ body {
...
+				font-size: 90%;
+				color: $code-color;
+				background-color: $code-background;
+				border: 1px solid $code-border;
+				border-radius: 3px;
+				padding-left: 2px;
```
drw158: Might as well write this as:

`padding: 0 2px;`

src/boot/stylesheets/main.scss
```diff
@@ -690,12 +690,41 @@ body {
     				display: block;
     			}
     		}
+
+			code {
+				font: $code;
+				font-size: 90%;
+				color: $code-color;
```
drw158: Nice, we don't really have a style for `code` in Calypso. Looks good, but with some color changes. First, I wouldn't use variables since this is a one-off deal and we try to keep the colors.scss file pretty clean.

Change the colors to this:

```
color: darken( $gray, 20% );
background: $gray-light;
border: 1px solid lighten( $gray, 20% );
```

src/boot/stylesheets/main.scss
dmsnell: Updated margin in 08a122e

src/boot/stylesheets/main.scss
dmsnell: adopted in 08a122e

src/boot/stylesheets/main.scss
dmsnell: compressed padding in 08a122e

src/boot/stylesheets/main.scss
dmsnell: inlined in 08a122e

dmsnell:
> Mentions seem to be broken, cutting off a lot of text:

@drw158 is there a way I can try to reproduce this. that seems odd and I want to make sure this code isn't doing something really funny but I'm not seeing the behavior/don't know if I'm seeing it

dmsnell:
Good call @drw158! in 798f875f5ba64a58c9e9f0e3fa926b994f2efa0f I fixed this behavior. it was a bug due to my misunderstanding of `String.split()` and has been fixed. I checked for other places this could have creeped in and I'm not using any other instances of that incorrectly.

Actually it was related to having a header-style paragraph with URLs (`:`s) in and not related to the mention.

src/templates/functions.jsx
```diff
@@ -22,28 +102,23 @@ export function p(html, className) {
...
-        var somewhatRandomKey =
-            'block-text-' + chunk.length + '-' + Date.now() + '-' + Math.random() * 10000;
+    return html.split('\n\n').map((chunk, i) => {
+        const key = `block-text-${i}-${chunk.length}-${chunk.slice(0, 3)}-${chunk.slice(-3)}`;
+        const blocks = toBlocks(chunk);
+
         return (
             <div
                 className={className}
-                key={somewhatRandomKey}
```
gwwar: 😄 `somewhatPsuedoRandomKey`

src/templates/functions.jsx
```diff
@@ -2,16 +2,96 @@ import React from 'react';
...
+        // Header: Value
+        //
+        // Not! This is fun. This: again; isn't emphasized.
+        // May detect false positive if colon found in first sentence.
+        const defined = /^[\w\s-_]+:/.test(raw)
+            ? `<strong>${raw.split(':')[0]}:</strong>${raw.replace(/^[^:]:+/, '')}`
+            : raw;
+
+        // inline `code` snippets
+        const coded = defined.replace(/`([^`]+)`/, '<code>$1</code>');
```
gwwar: I noticed a minor corner case with a `-` in a code block. I think we can probably ignore this for now, but might be worth considering constructing a proper tokenizer later if we need to handle more difficult cases in the future.
```
`- foo/bar`
```
screen shot 2017-05-31 at 3 39 12 pm

dmsnell: @gwwar I absolutely want to do this properly. it's basically only because I'm tired of seeing such unstyled content that I did this as a "meh, it's much better than it was"

I think you are right though in that inline code should have precedence over a list and I'll make this change.

having had such major issues in the past held me back from making some of these smaller improvements. when we "do it properly" (and why I hesitate from just doing that now) I want to make a drastic change in how we do all of the transformation. right now we pump API responses directly into the app and transform the data on every render. the approach I took in [dmsnell/wrnc](https://github.com/dmsnell/wrnc) was to [transform the API data when it comes into the system](https://github.com/dmsnell/wrnc/blob/58fe073c1c0b196ee46d02eb8dc178fbdc3c4227/src/api-poller/from-api.js) then we never have to repeat the transformation again. much faster, much saner. (data models, who would'a thunk it?)

dmsnell: @gwwar I can't reproduce. are we sure that isn't actually a bullet followed by inline code? (inline code already has precedence over lists)

can you share the underlying text causing this issue?

src/boot/stylesheets/main.scss
```diff
@@ -690,14 +690,53 @@ body {
...
+				background: $gray-light;
+				border: 1px solid lighten( $gray, 20% );
+				border-radius: 3px;
+				padding: 4px;
+
+				code {
+					border: none;
+					background: none;
+				}
+			}
```
dmsnell: @drw158 I added this section here because I was getting funny styling with `<code><pre>`

screen shot 2017-05-31 at 9 58 42 pm

I would be very grateful if you reviewed this here CSS…

src/templates/functions.jsx
gwwar: It's a code block over multiple lines eg
```
<code>
- foo
- bar
- baz
</code>
```

See p4TIVU-6vp-p2#comment-7105

dmsnell:
Is it worth not pushing this that the mobile clients don't support it?

drw158:
I think it's worth pushing even if mobile clients don't support it.

PR #115, opened 2017-05-29
Use Flexbox for vertical centering.

### LINKED ISSUE #113: Calypso: Vertically centered views look off
The calculations for centering things vertically aren't working as well in Calypso, as they seem to be off by the height of the master bar:

screen shot 2017-05-24 at 9 37 22 am

We'll need to be careful at fixing this so that it looks good in both standalone and in Calypso. Maybe we're using the height of the window, and should use the main div's height instead.

kwight:
Rather than using [JavaScript](https://github.com/Automattic/notifications-panel/blob/7c1815301b2af437a4fd6e918c96747cd99c0f0a/src/templates/empty-message.jsx#L56) and [math](https://github.com/Automattic/notifications-panel/blob/master/src/boot/stylesheets/main.scss#L306) to vertically center elements, this PR uses Flexbox.

screen shot 2017-05-29 at 2 55 26 pm
_Current / With fix / With fix + transform -50%. Note the environment badges were removed for clarity._

Using Flexbox still looks a little visually low (can't tell why), but adding back the transform that I originally pulled out seems to look better. SInce it's a `div` being vertically centered, I shouldn't _have_ to do that; but it does seem to be vertically aligned to the first line of text. @drw158 am I missing something? (Hm, maybe even `-25%` would be better.)

**Testing**
* Switch to this PR locally.
* Run the repo as standalone to be sure of no regressions.
* In both `packages.json` and `npm-shrinkwrap.json`, change the `"notifications-panel"` value to `"Automattic/notifications-panel#fix\/vertical-centering"`.
* `make distclean` and `make run`. Come back tomorrow.
* Check the notifications > Unread tab for spinner and message positioning.

Fixes #113.

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -323,9 +323,11 @@ body {
     }
 
     .wpnc__loading-indicator {
-    	display: block;
-    	background-color: $gray-light;
-    	height: 90px;
+		display: flex;
+		align-items: center;
+		justify-content: center;
+		background-color: $gray-light;
+		height: 100%;
     }
 
     .wpnc__note-list {
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -309,8 +309,9 @@ body {
     }
 
     .wpnc__empty-notes {
-    	text-align: center;
-    	padding: 0 32px;
+		text-align: center;
+		padding: 0 32px;
+		transform: translateY(-50%);
 
     	h2 {
     		font: 300 21px/24px $sans;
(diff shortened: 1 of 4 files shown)


drw158:
I think it's something to do with the 100% height. If you put in a px value, flex centers it perfectly.

If you go up the tree, you can see that the parent div `wpnc__list-view` is 100% height, and _its_ parent is `wpnc__note-list` which is the entire height of the whole panel, including the segmented control at the top. But I'm not sure... I think that would make the empty message appear a little _higher_ not lower.

Would be nice to sort this out, but I think this is a fine quick fix.

kwight:
> But I'm not sure... I think that would make the empty message appear a little higher not lower.

Right :/

I'll go ahead with this, and we can revisit later if need be.

PR #120, opened 2017-05-30
Add filter name to redux store

roundhill:
In order to add a solution for #103 we need to add the name of the currently selected filter to the redux store. This PR aims to accomplish that. 

I only stored the string value of the filter name in the store, but another option is to store the [entire Filter function](https://github.com/Automattic/notifications-panel/blob/master/src/templates/filters.js#L8-L28) for the current selection as well. Forgive me for any redux atrocities committed, still pretty new at this!

**To Test**
* Load the client fresh, you should arrive at the `All` filter by default.
* Changing filters should filter the list just like it used to.

### DIFF
--- a/src/state/ui/reducer.js
+++ b/src/state/ui/reducer.js
@@ -35,9 +35,8 @@ export const selectedNoteId = (state = null, { type, noteId }) => {
     return state;
 };
 
-export const filterName = (state = 'all', { type, filterName }) => {
-    return SET_FILTER === type ? filterName : state;
-};
+export const filterName = (state = 'all', { type, filterName }) =>
+    ( SET_FILTER === type ? filterName : state );
 
 export default combineReducers({
     isLoading,
--- a/src/state/ui/reducer.js
+++ b/src/state/ui/reducer.js
@@ -1,6 +1,12 @@
 import { combineReducers } from 'redux';
 
-import { NOTES_LOADED, NOTES_LOADING, SELECT_NOTE, SET_IS_SHOWING } from '../action-types';
+import {
+    NOTES_LOADED,
+    NOTES_LOADING,
+    SELECT_NOTE,
+    SET_IS_SHOWING,
+    SET_FILTER
+} from '../action-types';
 
 export const isLoading = (state = true, { type }) => {
     if (NOTES_LOADING === type) {
@@ -20,8 +26,13 @@ export const isPanelOpen = (state = false, { type, isShowing }) =>
 export const selectedNoteId = (state = null, { type, noteId }) =>
     (SELECT_NOTE === type ? noteId : state);
 
+export const filterName = (state = 'all', { type, filterName }) => {
+    return SET_FILTER === type ? filterName : state;
+};
+
 export default combineReducers({
     isLoading,
     isPanelOpen,
     selectedNoteId,
+    filterName,
 });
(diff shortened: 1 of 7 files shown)


src/templates/filter-bar-controller.js
```diff
@@ -31,12 +30,12 @@ FilterBarController.prototype.selectFilter = function(filterName) {
 };
 
 FilterBarController.prototype.getFilteredNotes = function(notes) {
-    if (!notes) {
+    const activeTab = Filters[store.getState().ui.filterName];
```
dmsnell: didn't you create a selector for this?

src/state/ui/reducer.js
```diff
@@ -20,8 +26,13 @@ export const isPanelOpen = (state = false, { type, isShowing }) =>
 export const selectedNoteId = (state = null, { type, noteId }) =>
     (SELECT_NOTE === type ? noteId : state);
 
+export const filterName = (state = 'all', { type, filterName }) => {
+    return SET_FILTER === type ? filterName : state;
```
dmsnell: I'm 100% okay with this, but it does break the pattern from its neighbors and doesn't have too. either way is great.

```js
export const filterName = ( state = 'all', { type, filterName } ) =>
	(SET_FILTER === type ? filterName : state);
```

src/templates/filter-bar-controller.js
```diff
@@ -19,7 +18,7 @@ FilterBarController.prototype.selectFilter = function(filterName) {
         return;
     }
 
-    this.selected = Filters[filterName]();
+    store.dispatch(actions.ui.setFilter(filterName));
```
dmsnell: if we're going to be saying that "changing the filter means unselecting the note" and we now have a filter change action, we should let this filter action invoke the change in the selected note and remove the call a few lines down to `actions.ui.unselectNote()`, instead handling it in the `selectedNote` reducer

dmsnell:
@roundhill I went ahead and committed here because I thought it might be educational. take a look at how the `selectedNote` reducer changed in 7cde77f. This is a good place to see how Redux actions and reducers can decouple actions from their effects/data changes and it can make for an easy audit of "why did this value change"

roundhill:
> I went ahead and committed here because I thought it might be educational. take a look at how the selectedNote reducer changed in 7cde77f.

Nice, thanks for the example.

src/templates/filter-bar.jsx
```diff
@@ -61,4 +63,8 @@ export const FilterBar = React.createClass({
     },
 });
 
-export default FilterBar;
+const mapStateToProps = state => ({
+    filterName: getFilterName(state),
+});
+
+export default connect(mapStateToProps, null)(FilterBar);
```
dmsnell: since `mapDispatchToProps` is optional, we don't need to pass `null` in, we can make it a single-arg function call

```js
connect( mapStateToProps )( FilterBar );
```

src/state/ui/reducer.js
roundhill: Nice catch, updated in ebdf4e8

src/templates/filter-bar.jsx
roundhill: Right, updated in bde9b9f!

roundhill:
Thanks for the review @dmsnell!

PR #121, opened 2017-05-31
Fix unread notes from disappearing when clicked while viewing the unread filter

roundhill:
Add a new array that tracks note reads, and when in the unread filter add these notes to the list so that they 'stick' even after you've read them while viewing the filter.

Changing filters will always reset the list.

**To test**
* View the unread filter when it has some unread notes. Clicking on a note should open the detail view, but not remove the note from the list. 
* When you change filters and go back to the unread filter, the notes you read should now be gone.

### DIFF
--- a/src/templates/filter-bar-controller.js
+++ b/src/templates/filter-bar-controller.js
@@ -1,3 +1,5 @@
+import { includes } from 'lodash';
+
 import Filters from './filters';
 import { store } from '../state';
 import actions from '../state/actions';
@@ -39,7 +41,7 @@ FilterBarController.prototype.getFilteredNotes = function(notes) {
 
     const filteredNoteReads = getFilteredNoteReads(state);
     const filterFunction = (note) =>
-        (filterName === 'unread' && filteredNoteReads.includes(note.id)) || activeTab().filter(note);
+        (filterName === 'unread' && includes(filteredNoteReads, note.id)) || activeTab().filter(note);
 
     return notes.filter(filterFunction);
 };
--- a/src/templates/filter-bar-controller.js
+++ b/src/templates/filter-bar-controller.js
@@ -1,4 +1,4 @@
-import { includes } from 'lodash';
+import { includes, some } from 'lodash';
 
 import Filters from './filters';
 import { store } from '../state';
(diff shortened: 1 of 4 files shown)


src/templates/filter-bar-controller.js
```diff
@@ -29,12 +29,17 @@ FilterBarController.prototype.selectFilter = function(filterName) {
...
 FilterBarController.prototype.getFilteredNotes = function(notes) {
-    const activeTab = Filters[getFilterName(store.getState())];
+    const filterName = getFilterName(store.getState());
+    const activeTab = Filters[filterName];
     if (!notes || !activeTab) {
         return [];
     }
 
-    return notes.filter(activeTab().filter);
+    const filteredNoteReads = store.getState().notes.filteredNoteReads;
```
dmsnell: Should be using a selector here

src/templates/filter-bar-controller.js
```diff
@@ -29,12 +29,17 @@ FilterBarController.prototype.selectFilter = function(filterName) {
...
+    const filterFunction = (note) =>
+        (filterName === 'unread' && filteredNoteReads.includes(note.id)) || activeTab().filter(note);
```
dmsnell: note that we have no IE support for `Array.includes()` so we should either use `lodash.includes()` or change the data structure

```js
if ( SELECT_NOTE )
	return { ...state, [ note.id ]: true };

…

noteHasFilteredRead( note.id )

…

noteHasFilteredRead = ( state, noteId ) => !! getNotesState( state ).filteredNotes[ noteId ];
```

roundhill:
> note that we have no IE support for Array.includes()

Oof! Glad you caught that.

roundhill:
Thanks for the tips @dmsnell. I've added the selector, and used lodash's `includes` and `some` to clean up the filtering code.

src/templates/filter-bar-controller.js
```diff
@@ -29,12 +32,20 @@ FilterBarController.prototype.selectFilter = function(filterName) {
...
-    const activeTab = Filters[getFilterName(store.getState())];
+    const state = store.getState();
+    const filterName = getFilterName(state);
+    const activeTab = Filters[filterName];
     if (!notes || !activeTab) {
         return [];
     }
 
-    return notes.filter(activeTab().filter);
+    const filteredNoteReads = getFilteredNoteReads(state);
```
dmsnell: do we ever need this list? what if it weren't stored as a list (as it probably won't always)

what if instead we had the selector return if it's a filtered read note?

```js
'unread' === filterName && isFilteredNoteRead( note.id )
``

roundhill: Yeah I saw your note(heh) about that in slack. We don't need the list anywhere else, so I'll add a selector for just checking if the note id is in the list 👍

roundhill: Done in c35ce52. Looking a lot better 👍

PR #124, opened 2017-06-01
Restore missing "empty" messages.

kwight:
**Testing**
* Go to the Unread filter and clear any unread messages. You will have a blank screen.
* Apply this PR, and do the same; you'll now have the "You're all caught up!" message.


roundhill:
I caused this regression in #120. 

Looks good! :shipit:

PR #126, opened 2017-06-01
Bring parity to iFrame version

dmsnell:
- Send close panel request with `n` key
 - Stop auto-opening panel after boot

### DIFF
--- a/src/Notifications.jsx
+++ b/src/Notifications.jsx
@@ -84,7 +84,6 @@ export class Notifications extends PureComponent {
          *
          * @TODO: Pass this information directly into the Redux initial state
          */
-        store.dispatch(isShowing ? actions.ui.openPanel() : actions.ui.closePanel());
         store.dispatch({ type: SET_IS_SHOWING, isShowing });
 
         client.setVisibility({ isShowing, isVisible });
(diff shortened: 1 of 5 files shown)


kwight:
Curious what this is for? I can already use `n` to toggle the notifications panel open and closed.

I think this may also introduce a regression; `enter` on a note now closes the panel, rather than opening the single view.

dmsnell:
> I think this may also introduce a regression; enter on a note now closes the panel, rather than opening the single view.

can you list reproduction steps @kwight or capture it in a video? I'm not seeing it.

> Curious what this is for? I can already use n to toggle the notifications panel open and closed.

in the iFrame version I'm able to open the panel with `n` but not close it. this patch should fix that

src/Notifications.jsx
```diff
@@ -95,22 +94,17 @@ export class Notifications extends PureComponent {
     }
 
     componentWillReceiveProps({ isShowing, isVisible, wpcom }) {
-        initAPI(wpcom);
-
-        if (this.props.isShowing && !isShowing) {
-            store.dispatch(actions.ui.closePanel());
-            // unselect the note so keyhandlers don't steal keystrokes
-            reset();
```
gwwar: With our current implementation we still need this since we will steal keystrokes if a note is left open and we close the panel. See: https://github.com/Automattic/notifications-panel/pull/106

Basically open a post/comment note, close the panel. We won't be able to type `l` or any of the other bound keys in the Calypso editor/text fields.

dmsnell: good call! added back in in e0c7979

I left the `closePanel()` call out of it though

PR #132, opened 2017-06-05
Replace anchor markup with <button> for action buttons.

kwight:
This prevents event propagation when an action is clicked, which can trigger a "You have unsaved changes" browser alert.

Also fixes a styling bug where the text and gridicons of actions would be different colors on hover.

screen shot 2017-06-05 at 12 56 05 pm
_Before_
screen shot 2017-06-05 at 12 55 51 pm
_After_

Fixes https://github.com/Automattic/wp-calypso/issues/14071.

### DIFF
--- a/src/boot/stylesheets/actions.scss
+++ b/src/boot/stylesheets/actions.scss
@@ -1,4 +1,8 @@
 .wpnc__note-actions {
+	.wpnc__note-actions__buttons {
+		padding-bottom: 16px;
+	}
+
 	.wpnc__reply-box {
 		font-family: $sans;
 		display: block;
@@ -53,12 +57,49 @@
 	}
 }
 
-.wpnc__note-actions__buttons div.wpnc__action-link {
+.wpnc__action-link {
+	background-color: transparent;
+	border: none;
+	box-sizing: border-box;
+	cursor: pointer;
 	display: inline-block;
-	vertical-align: top;
+	font-size: 12px;
 	max-width: 25%;
 	padding: 0 18px;
-	box-sizing: border-box;
+	text-align: center;
+	vertical-align: top;
+
+	&.active-action {
+		color: $orange-jazzy;
+
+		&:hover {
+			color: $gray-text-min;
+
+			.gridicon {
+				fill: $gray;
+			}
+		}
+
+		.gridicon {
+			fill: $orange-jazzy;
+		}
+	}
+
+	&.inactive-action {
+		color: $gray-text-min;
+
+		&:hover {
+			color: $orange-jazzy;
+
+			.gridicon {
+				fill: $orange-jazzy;
+			}
+		}
+
+		.gridicon {
+			fill: $gray;
+		}
+	}
 
 	@media screen and (max-width: 370px) {
 		padding: 0 12px;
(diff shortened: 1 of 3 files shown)


src/boot/stylesheets/actions.scss
```diff
@@ -75,34 +75,44 @@
...
-	a.active-action:hover {
-		color: $gray-text-min;
-	}
 }
 
 .wpnc__main div.wpnc__action-link {
+	cursor: pointer;
 	text-align: center;
 	font-size: 12px;
+	height: 55px;
```
gwwar: Curious, but why do we need to set height here? Did we used to have a default height on links?

src/boot/stylesheets/actions.scss
```diff
@@ -75,34 +75,44 @@
...
 
 	.wpnc__noticon {
 		font-size: 24px;
 	}
 
-	a.active-action {
+	.active-action {
```
gwwar: The existing modifier classes are fine, but just a heads up that we'll want to namespace our classes with `wpnc` for new class names.

src/templates/action-button.jsx
```diff
@@ -25,14 +25,13 @@ const ActionButton = React.createClass({
         return (
             <HotkeyContainer shortcuts={hotkeys}>
                 <div className="wpnc__action-link">
-                    <a
-                        href="#"
+                    <div
```
gwwar: Do we need an extra container div, or can we move these props to the parent?

drw158:
What about using `button`?

gwwar:
> What about using button?

Seems like that would probably make more semantic sense than a div 👍

kwight:
> What about using button?

@drw158 I took a look through some existing things, like our filter (uses an `ul`), and the Reader's similar actions (uses `div`s within an `ul`). It seemed like a reasonable choice, but `button` would work too. For consistency, maybe the best choice is another `ul`?..

kwight:
> Basically, the icon should be $gray and text should be $gray-text-min. On hover, both of them need to be orange.

Ah, got it – fixed in 08671b50709a76d82c4aa894c8bc476943ab7f1f.

src/boot/stylesheets/actions.scss
kwight: Busted – I couldn't determine the reason for the effective height of 55px, which comes from `wpnc__note-actions__buttons` – I'll dig deeper.

src/templates/action-button.jsx
kwight: Good point – I'll see what that's like.

gwwar: Note that if the top level changes to a Button, we'll likely want to update the display from `inline-block` to `block`

src/boot/stylesheets/actions.scss
kwight: Removed and no longer applicable with the refactor in 3cdefcff4dab7f3d899bac5b1823cf8c5c280806 and 00064b2c268b07d248188746b5144a6c91d198eb.

kwight:
> Do we need an extra container div, or can we move these props to the parent?

Container div removed in 3cdefcf and 00064b2. Technically, we could further eliminate another wrapper (`wpnc__note-actions__buttons`), but then we need to get a little hacky and aggressive to handle the bottom margins (and `:last-child` padding of the Spam button). Feels better to leave that one in place.

dmsnell:
I'm all for `<button>` or `<a>` over `<div>`. If the evens are propagating we can handle that just as easily too with `event.preventDefault()` and `event.stopPropagation()`

PR #133, opened 2017-06-06
Remove back arrow noticon

drw158:
Replaced the noticon with a gridicon back arrow.

Also removed some duplicated svg code.

Before:

screen shot 2017-06-06 at 10 53 32 am

After:

screen shot 2017-06-06 at 10 53 12 am

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -76,12 +76,10 @@ body {
     		display: inline;
     		float: left;
 
-    		&:before {
-    			@extend %wpnc__noticon;
-    			content: '\f431';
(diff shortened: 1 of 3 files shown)


gwwar:
> the wrapping <a> could be changed to a button

I feel that links should mostly be used for navigation and buttons for actions. In this case I think a link is a better fit, and we can avoid the anchor issue by calling `event.preventDefault()` in the onClick handler.

kwight:
@gwwar so navigation with `<a>`s can be internal (the back, up, or down links) or external (links to comment single views), while actions are `<button`s – awesome, wfm 👍

PR #134, opened 2017-06-06
Remove nav arrow noticons

drw158:
I couldn't figure out how to do the up/down icon switch in the component itself. Had to just flip the icon in the CSS. It might help me to know how to do it in the component though. Any guidance is appreciated!

Before:

screen shot 2017-06-06 at 11 55 23 am

After:

screen shot 2017-06-06 at 11 55 07 am

### DIFF
--- a/src/templates/nav-button.jsx
+++ b/src/templates/nav-button.jsx
@@ -1,5 +1,13 @@
+/**
+ * External dependencies
+ */
 import React, { Component, PropTypes } from 'react';
 import classNames from 'classnames';
+
+/**
+ * Internal dependencies
+ */
+import Gridicon from './gridicons';
 
 export const NavButton = React.createClass({
     navigate(event) {
@@ -18,7 +26,10 @@ export const NavButton = React.createClass({
                 disabled={!this.props.isEnabled}
                 href="#"
                 onClick={this.props.isEnabled ? this.navigate : null}
-            />
+            >
+              <Gridicon icon="arrow-down" size={18} />
+            </a>
+
         );
     },
 });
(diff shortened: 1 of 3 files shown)


src/templates/nav-button.jsx
```diff
@@ -18,7 +26,10 @@ export const NavButton = React.createClass({
                 disabled={!this.props.isEnabled}
                 href="#"
                 onClick={this.props.isEnabled ? this.navigate : null}
-            />
+            >
+              <Gridicon icon="arrow-down" size={18} />
```
gwwar: It looks like we do have an `arrow-up` gridicon as well. I'd prefer if we used it directly since a `transform` may make the icon look a bit blurry.

We have a few options here, we can toggle based off the className, or add an explicit prop. I prefer the latter.

So this might look like
```jsxt
  <NavButton
       className="wpnc__prev"
       iconName="arrow-up"
       //...
  />
  <NavButton
       className="wpnc__next"
       iconName="arrow-down"
       //...
  />
```

Where in this component we do:
```jsx
<Gridicon icon={ this.props.iconName } size={ 18 } />
```

gwwar: Oh and don't forget to add it to the propTypes:
```jsx
NavButton.propTypes = {
    iconName: PropTypes.string.isRequired,
    //...
}
```

gwwar:
Overall the controls work as expected 👍. Ping me again when you're ready for another 👀

drw158:
@gwwar  I should've thought of that! Thanks, ready for another review.

PR #135, opened 2017-06-06
Tidy up event propagation of anchors.

kwight:
Noting our distinction of [when to use anchors vs. button elements](https://github.com/Automattic/notifications-panel/pull/133#issuecomment-306546364), this PR tidies up some unnecessary `href` attributes, while ensuring no event bubbling when clicking the Back button. This also guards more anchors against the original issue in Automattic/wp-calypso#14071.

### DIFF
--- a/src/templates/button-back.jsx
+++ b/src/templates/button-back.jsx
@@ -13,6 +13,7 @@ import actions from '../state/actions';
 import Gridicon from './gridicons';
 
 const routeBack = (global, unselectNote) => event => {
+    event.stopPropagation();
     event.preventDefault();
     global.input.lastInputWasKeyboard = false;
     unselectNote();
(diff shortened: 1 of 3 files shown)


kwight:
@dmsnell The `href` attributes are [optional](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) and don't serve a purpose in these cases, being only internal UI changes. Removing the `href` also means no confusing "linking to" notification in browsers with a trailing hash on mouseover:
screen shot 2017-06-06 at 1 31 07 pm
_Mousing over Follow_

dmsnell:
Cool!

kwight:
@dmsnell Sweet, is this PR good to go for you, then?

PR #136, opened 2017-06-06
Use button elements instead of anchors for actions.

kwight:
Noting that we want to use [anchors for navigation and buttons for actions](https://github.com/Automattic/notifications-panel/pull/133#issuecomment-306546364), this PR corrects any instances of anchors that should instead be buttons.

The following three visual elements are affected:

screen shot 2017-06-06 at 12 48 15 pm
_Comment reply SEND button_

screen shot 2017-06-06 at 12 49 31 pm
_Follow button_

screen shot 2017-06-06 at 12 50 27 pm
screen shot 2017-06-06 at 12 51 20 pm
_The Trash and Spam actions_

**Testing**
* Open both a notifications standalone window with this PR, and a WordPress.com tab.
* Compare the three elements above, and be sure there are no functional or styling regressions.

### DIFF
--- a/src/boot/stylesheets/actions.scss
+++ b/src/boot/stylesheets/actions.scss
@@ -41,6 +41,7 @@
 			font-size: 12px;
 		}
 		button.active {
+			display: block;
 			color: $blue-medium;
 		}
 		button.inactive {
--- a/src/boot/stylesheets/actions.scss
+++ b/src/boot/stylesheets/actions.scss
@@ -56,7 +56,7 @@
 	}
 }
 
-.wpnc__action-link {
+.wpnc__main .wpnc__action-link {
 	background-color: transparent;
 	border: none;
 	box-sizing: border-box;
--- a/src/boot/stylesheets/actions.scss
+++ b/src/boot/stylesheets/actions.scss
@@ -25,11 +25,11 @@
 			resize: vertical;
 		}
 
-		textarea:focus + a {
+		textarea:focus + button {
 			display: block;
 		}
 
-		a {
+		button {
 			font-family: $sans;
 			color: $gray;
 			display: none;
@@ -40,11 +40,10 @@
 			font-weight: 600;
 			font-size: 12px;
 		}
-		a.active {
-			display: block;
+		button.active {
 			color: $blue-medium;
 		}
-		a.inactive {
+		button.inactive {
 			cursor: default;
 		}
 
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -44,6 +44,18 @@ body {
     	color: $gray;
     }
 
+    button {
+		background-color: transparent;
+		border: none;
+		cursor: pointer;
+		font-size: inherit;
+		padding: 0;
+
+		&.focus {
+			outline: none;
+		}
+	}
+
     header {
     	border-bottom: 1px solid lighten( $gray, 30 );
     	box-sizing: border-box;
--- a/src/templates/comment-reply-input.jsx
+++ b/src/templates/comment-reply-input.jsx
@@ -295,24 +295,23 @@ const CommentReplyInput = React.createClass({
                     context: 'verb: imperative',
                 });
                 submitLink = (
-                    <a
-                        href="#"
+                    <button
                         title={submitLinkTitle}
                         className="active"
                         onClick={this.handleSubmit}
                         onKeyDown={this.handleSendEnter}
                     >
                         {sendText}
-                    </a>
+                    </button>
                 );
             } else {
                 var submitLinkTitle = this.props.translate(
                     'Write your response in order to submit'
                 );
                 submitLink = (
-                    <a href="#" title={submitLinkTitle} className="inactive">
+                    <button title={submitLinkTitle} className="inactive">
                         {sendText}
-                    </a>
+                    </button>
                 );
             }
         }
--- a/src/templates/follow-link.jsx
+++ b/src/templates/follow-link.jsx
@@ -74,10 +74,10 @@ export const FollowLink = React.createClass({
         }
 
         return (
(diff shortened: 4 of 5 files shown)


drw158:
Maybe the close links should also be `button`? Action?

The SEND action does not work for me. After clicking SEND, nothing happens, and the loading spinner does not appear.

src/templates/comment-reply-input.jsx
```diff
@@ -295,24 +295,23 @@ const CommentReplyInput = React.createClass({
                     context: 'verb: imperative',
                 });
                 submitLink = (
```
gwwar: I think we'll want to update these variable names as well, otherwise it may be a bit confusing.

src/boot/stylesheets/main.scss
```diff
@@ -44,6 +44,18 @@ body {
     	color: $gray;
     }
 
+    button {
+		background-color: transparent;
```
gwwar: I guess how do we feel about links that perform actions?

While it's not ideal, it also feels strange that we're setting some base styling for buttons to look like links.

kwight:
> The SEND action does not work for me. After clicking SEND, nothing happens, and the loading spinner does not appear.

Doh, put that in the "now why is this CSS property here?.." file – fixed in d0d35fb585373211c839dd918d00c8c4d5e06eb2.

> Maybe the close links should also be button? Action?

Hm, missed those because they're `<span>`s; yes, for consistency, they'd be good as button elements too. I've taken a stab at them, but they have other issues; I'll tackle that in another PR, as this is already getting big.

src/boot/stylesheets/main.scss
kwight: > I guess how do we feel about links that perform actions?

Is your definition of a "link" when the trigger is just text? I guess I've always thought that a button triggers an action (being just text or not), while an anchor/link sends us somewhere in the browser.

> While it's not ideal, it also feels strange that we're setting some base styling for buttons to look like links.

The unfortunate thing about the `<button>` element is that it's heavily styled by default in browsers, and seems like it always needs a certain amount of reset:

screen shot 2017-06-06 at 3 13 46 pm

gwwar: > Is your definition of a "link" when the trigger is just text?

That would be my first thought when scanning a page. Though it can be sometimes be used as an action when we really don't a person to do a particular thing. For example having a button that looks like a button as the primary and a link `cancel` as the secondary action. 

I usually defer to whatever our common design language happens to be, but I was curious on thoughts about this.

src/templates/comment-reply-input.jsx
gwwar: Maybe follow up on this in another PR?

src/boot/stylesheets/actions.scss
```diff
@@ -57,7 +57,7 @@
 	}
 }
 
-.wpnc__action-link {
+.wpnc__main .wpnc__action-link {
```
gwwar: Did we need the extra CSS specificity here?

drw158:
I like to think of WP as an app. So, unless it's a link with a meaningful href (www.google.com), it should probably be a button. That being said I'm not sure what we do in Calypso.

src/boot/stylesheets/actions.scss
kwight: We do, unfortunately, to override the `.wpnc__main button` in `main.scss`. I put the `button` declaration within `wpnc__main` in `main.scss` because every other element styling was done the same way (except for `body`) – _feels_ safer (and more consistent)?

I'll go ahead with this, and we can revisit later if we want to re-evaluate `main.scss`.

kwight:
> That being said I'm not sure what we do in Calypso.

Calypso seems to be pretty consistent with using anchors for anything that sends us to a valid URL, and buttons for anything else; for instance, all of these (including the back arrow) are buttons:

screen shot 2017-06-07 at 10 31 53 am

There's one notable exception: the notifications bell. It's an anchor (like everything else in the masterbar), but it's the only item that doesn't result in a full layout change, with a new valid URL. It's even kind of confusing; mousing over the bell implies we have a `/notifications` route, which is invalid and gives a 404 page if entered directly into the address bar. I think I'll do a Calypso issue for that.

src/boot/stylesheets/actions.scss
gwwar: The general resets do need to go under `.wpnc__main` otherwise Calypso will pick up those rules as well. 😄 

Depending on the order of declaration `button.wpnc__action-link` might also work, but this is fine here too.

kwight: > The general resets do need to go under .wpnc__main otherwise Calypso will pick up those rules as well

Of course! I knew it felt funny for a reason 🙃 

I think it _did_ work, but then I saw "Avoid using over-qualified selectors like div.my-class" in the [Sass coding guidelines](https://wpcalypso.wordpress.com/devdocs/docs/coding-guidelines/css.md). I admit I haven't totally grokked those guidelines as a whole (and have never worked with Sass before 😱), so I just went with this.

gwwar: For CSS specificity, it's just counting. ids always beat classes, classes always beat elements. So say if you have a selector with 100 elements and a second with one class, the second selector will win.

eg `( ids, classes, elements )` with inline styles or `!important` overruling any other rules.

In Calypso we try for ( 0, 1, 0 ) and for the most part use Component based rules than global resets, to avoid specificity issues like the one here.

In this example `button.wpnc__action-link` is ( 0, 1, 1 ). 
whiele `.wpnc__main .wpnc__action-link` is ( 0, 2, 0 ).

Which means that the second rule here has a higher specificity. It's not a huge deal here, but I like to trim down specificity if we don't need it.

kwight:
> Maybe the close links should also be button? Action?

Addressed in #140 .

PR #137, opened 2017-06-06
Replace cross noticon

drw158:
Replaced noticon for the undo message. To test, delete a comment.

Replaced cross noticon for the status message. To test, reply to a comment.

Before:

screen shot 2017-06-06 at 2 55 43 pm
screen shot 2017-06-06 at 2 58 36 pm

After:

screen shot 2017-06-06 at 2 55 28 pm
screen shot 2017-06-06 at 2 58 52 pm

### DIFF
--- a/src/templates/undo-list-item.jsx
+++ b/src/templates/undo-list-item.jsx
@@ -9,6 +9,8 @@ import getSelectedNoteId from '../state/selectors/get-selected-note-id';
 
 import { bumpStat } from '../rest-client/bump-stat';
 
+import Gridicon from './gridicons';
+
 var { recordTracksEvent } = require('../helpers/stats');
 
 const KEY_U = 85;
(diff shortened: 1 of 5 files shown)


src/templates/undo-list-item.jsx
```diff
@@ -195,7 +197,7 @@ export const UndoListItem = React.createClass({
                     </a>
                     <span className="wpnc__undo-message">{message}</span>
                     <span className="wpnc__close-link wpnc__noticon" onClick={this.actImmediately}>
```
kwight: This `wpnc__noticon` class can be removed.

drw158:
Thanks, all done. Also removed the unneeded css there.

kwight:
👍 

Side note, it looks like the cross has slid to the left, compared to your screenshot in the description above (it's also to the left in prod)?

screen shot 2017-06-06 at 2 03 55 pm

drw158:
Hm, I don't see the X icon being that far to the left in my environments. What browser and environment is that in?

kwight:
Chrome 59 + macOS Sierra.

PR #140, opened 2017-06-07
Replace close link anchors with button elements.

kwight:
Following from #132 , this replaces the close `X` anchors with buttons. There are further markup differences between the two instances (the overlay uses `flex`, while the other does not), but they could be better dealt with in a separate PR.

This PR also addresses a visual bug where the overlay notice doesn't extend the full width of its container:

screen shot 2017-06-07 at 12 17 29 pm
_Before_

screen shot 2017-06-07 at 12 51 26 pm
_After_

### DIFF
--- a/src/boot/stylesheets/overlay-bars.scss
+++ b/src/boot/stylesheets/overlay-bars.scss
@@ -9,10 +9,6 @@
 	background-color: $alert-red;
 	color: $white;
 
-  @media only screen and (min-width: 800px) {
-		width: 400px;
-	}
-
 	a,
 	a:visited {
 		color: $white;
@@ -26,17 +22,13 @@
 		color: $white;
 	}
 
-	.wpnc__status-bar__wpnc__close-link {
-		display: inline-block;
-
-		.gridicon {
-			vertical-align: middle;
-		}
+	span {
+		margin: 15px 5px 15px 5px;
 	}
 
-  span {
-		margin: 15px 5px 15px 5px;
-  }
+	.wpnc__close-link {
+		position: relative;
+	}
 }
 
 .wpnc__status-bar.success {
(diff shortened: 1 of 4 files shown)


drw158:
Looks good to me. Sort of weird that the notice overlays the scrollbar (if they are turned on).

I don't mind it though.

screen shot 2017-06-07 at 3 35 24 pm

src/boot/stylesheets/overlay-bars.scss
```diff
@@ -9,10 +9,6 @@
 	background-color: $alert-red;
 	color: $white;
 
-  @media only screen and (min-width: 800px) {
-		width: 400px;
```
gwwar: Is it ok to remove this? In small viewports I don't see the green notice.

gwwar: Oh weird it shows up once you leave the view, I guess this is fine then 😄 but we might want to revisit later

kwight: I believe that's a separate issue; you need to click the Back button to notice it on the main list view – it's the same in prod right now. I meant to ask @drw158 if there was any reason that was a deliberate choice, or just an oversight for the narrow single view.

drw158: Not sure why we did that. So long ago... 👴

kwight:
> Sort of weird that the notice overlays the scrollbar (if they are turned on).

@drw158 Oh, interesting; I always have them on "magic" mode. This way it matches the width of the tabs directly above it, and doesn't display a gap in "when scrolling" mode, which I believe is the default. Since you're OK with it, let's merge this and we can look at it in another PR going forward.

PR #141, opened 2017-06-07
Remove noticon extends

drw158:
This PR removes the use of the noticon extend, which previously provided an easy way to style the noticon icon font.

Should be no visual change, other than maybe a 1px difference.

Some elements depended on some of the noticon extend styles though:

```
	display: inline-block;
	font: normal 16px/1 'Noticons';
	vertical-align: top;
```

I went through each use of the extend and made sure it looked the same after removal. For most things, I had to add some styles back in, like vertical align.

Most notable changes:

- icon bubble
- 'has reply' icons

screen shot 2017-06-07 at 3 32 43 pm

screen shot 2017-06-07 at 3 32 58 pm

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -397,7 +397,8 @@ body {
     		}
 
     		.wpnc__subject .wpnc__noticon {
-    			@extend %wpnc__noticon;
+    			line-height: 1;
+					vertical-align: -3px;
     			color: $gray;
     			padding: 2px 5px 0 0;
     		}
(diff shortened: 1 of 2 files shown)


src/boot/stylesheets/main.scss
```diff
@@ -267,10 +261,12 @@ body {
     	}
 
     	.wpnc__note-icon .wpnc__noticon {
-    		@extend %wpnc__noticon;
```
gwwar: ✨

gwwar:
For folks 👀, it'd probably be good to test this in a rtl language too.

kwight:
Tested both LTR and RTL as a branch in Calypso; looks great, works well 👍

dmsnell:
Did we test this in the iframe too?

drw158:
I only tested the local build. It _shouldn't_ be any different in the iframe.

dmsnell:
I asked just because the stylesheets get transformed in the iframe version too

PR #144, opened 2017-06-08
Fix: Note reads across browser sessions, <Note> cleanup

### LINKED ISSUE #130: `Read` state not syncing to multiple browser tabs
**To reproduce**
1. Open two new calypso tabs with the notifications panel open.
2. Mark a notification as read in one tab.
3. Switch to the other tab, and observe that the notification is still unread.

Note: Changing tabs in the second tab appears to update the unread state, so maybe we just need to force a `render()`?

cc @dmsnell for any thoughts.

dmsnell:
Resolves #130

In the earlier versions of the app we used stateful updates to the React
render and relied on storage events to communicate between sessions that
notes were being read. This led to re-renders reflecting those changes
by listening to the `ready` event from `rest-client`.

With the introduction of Redux and the removal of those event-emitters
we lost this functionality and notes weren't being marked read. Further,
because we were still tied to the _impure_ render modes the note
component wasn't updating even when state was.

Now, upon retrieval of such a storage message we are marking the note as
read via a dispatched action. The relevant components are now no longer
marked as _impure_ and will update on all state changes.

Interestingly enough an old `shouldComponentUpdate` functionw was
preventing updates because it wasn't comparing against `getNoteIsRead()`
(since it's now separate from the note object).

**Testing**

See #130 - open two browser tabs/sessions and read a note in one of them. In **master** it should not carry over but in this branch it should update practically instantly in both or all sessions.

### DIFF
--- a/src/rest-client/index.js
+++ b/src/rest-client/index.js
@@ -1,7 +1,7 @@
 /**
  * Module dependencies.
  */
-import { difference, get, max, pick, property, range } from 'lodash';
+import { difference, get, pick, property, range } from 'lodash';
 
 import { store } from '../state';
 import actions from '../state/actions';
@@ -459,6 +459,13 @@ function handleStorageEvent(event) {
                 });
             }
         } catch (e) {}
+        return;
+    }
+
+    if ('note_read_status_' === event.key.substring(0, 17)) {
+        const noteId = parseInt(event.key.slice(17), 10);
+
+        return store.dispatch(actions.notes.readNote(noteId));
     }
 }
 
--- a/src/templates/note.jsx
+++ b/src/templates/note.jsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { connect } from 'react-redux';
+import classNames from 'classnames';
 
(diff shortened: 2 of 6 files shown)


src/rest-client/index.js
```diff
@@ -459,6 +459,13 @@ function handleStorageEvent(event) {
                 });
             }
         } catch (e) {}
+        return;
+    }
+
+    if ('note_read_status_' === event.key.substring(0, 17)) {
+        const noteId = parseInt(event.key.slice(17), 10);
+
+        return store.dispatch(actions.notes.readNote(noteId));
```
roundhill: 👍

src/templates/note.jsx
```diff
@@ -10,104 +11,62 @@ import NoteBody from './body';
...
-            summary = (
-                <SummaryInList
-                    currentNote={this.props.currentNote}
-                    key={'note-summary-list' + this.props.note.id}
-                    note={this.props.note}
-                    global={this.props.global}
-                />
-            );
-        }
+    const classes = classNames('wpnc__note', `wpnc__${note.type}`, {
```
roundhill: 👍

PR #145, opened 2017-06-08
Rename class names to gridicons

drw158:
This finally removes noticons and the link to the noticons css file.

There is one last classname that I can't figure out `wpnc__noticon`

screen shot 2017-06-08 at 11 08 13 am

screen shot 2017-06-08 at 11 11 55 am

I believe it's being created here:

src/indices-to-html/index.js

but I can't make sense of the code.

### DIFF
--- a/src/indices-to-html/index.js
+++ b/src/indices-to-html/index.js
@@ -26,7 +26,7 @@ function render_range(new_sub_text, new_sub_range, range_info, range_data, optio
         type_mappings = {
             b: 'strong', // be strong, my friend
             i: 'em', // I am, don't worry
-			noticon: 'gridicon',
+            noticon: 'gridicon',
         };
 
         // Replace unwanted tags with more popular and cool ones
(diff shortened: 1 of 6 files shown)


kwight:
Looks like the Reply icon has lost some style and positioning:

screen shot 2017-06-08 at 10 45 50 am
_This PR_

screen shot 2017-06-08 at 10 46 28 am
_Master_

kwight:
I pushed a commit that fixes that class, but I'm a little leary about how I'm doing it; does this look OK @dmsnell ?

dmsnell:
@kwight this is fine. if we eventually make an API change (which would be good) it will need full support in all our clients before we do

please note the formatting and auto-formatting. I saw some issues and we will need to run `prettier` against the changed files.

drw158:
Ok, so only thing left to do is to run Prettier? I'm not sure how to do that.

src/indices-to-html/index.js
```diff
@@ -26,14 +26,15 @@ function render_range(new_sub_text, new_sub_range, range_info, range_data, optio
         type_mappings = {
             b: 'strong', // be strong, my friend
             i: 'em', // I am, don't worry
+			noticon: 'gridicon',
```
dmsnell: just noticed from this line that prettier hadn't appeared to run. we can just add a space

drw158:
Thanks for all the help!

PR #148, opened 2017-06-08
Update the single view header navigation markup.

kwight:
This PR uses `<button>` instead of anchors for the single view header navigation markup.

No visible regressions should be seen.

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -39,20 +39,18 @@ body {
     	color: $blue-wordpress;
     }
 
-    a[disabled], .wpnc__single-view header .back[disabled] {
-    	pointer-events: none;
-    	color: $gray;
-    }
-
     button {
 		background-color: transparent;
 		border: none;
+		color: $blue-wordpress;
 		cursor: pointer;
 		font-size: inherit;
+		outline: none;
 		padding: 0;
 
-		&.focus {
-			outline: none;
+		&[disabled] {
+			pointer-events: none;
+			color: $gray;
 		}
 	}
 
@@ -81,6 +79,10 @@ body {
     		color: darken( $gray, 10% );
     		display: inline;
     	}
+
+		button {
+			line-height: 38px;
+		}
 
     	.wpnc__back {
     		@extend %headertext;
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -49,8 +49,8 @@ body {
 		padding: 0;
 
 		&[disabled] {
-			pointer-events: none;
 			color: $gray;
+			cursor: default;
 		}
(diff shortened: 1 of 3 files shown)


src/boot/stylesheets/main.scss
```diff
@@ -39,20 +39,18 @@ body {
...
+		color: $blue-wordpress;
 		cursor: pointer;
 		font-size: inherit;
+		outline: none;
 		padding: 0;
 
-		&.focus {
-			outline: none;
+		&[disabled] {
+			pointer-events: none;
```
gwwar: Do we need this? Usually buttons are unclickable when it has a `disabled` prop.

src/templates/button-back.jsx
```diff
@@ -25,14 +25,14 @@ export const BackButton = ({ global, isEnabled, translate, unselectNote }) => {
...
 
     return isEnabled
-        ? <a className="wpnc__back" onClick={routeBack(global, unselectNote)}>
+        ? <button className="wpnc__back" onClick={routeBack(global, unselectNote)}>
               <Gridicon icon="arrow-left" size={18} />
               {backText}
-          </a>
-        : <a className="wpnc__back disabled" disabled="disabled">
+          </button>
+        : <button className="wpnc__back disabled" disabled="disabled">
```
gwwar: This can be `<button className="wpnc__back disabled" disabled>`. React is also smart enough to handle html5 boolean props correctly:

So this might change to:
```jsx
<button 
     className="wpnc__back" 
     onClick={routeBack(global, unselectNote)}
      disabled={!isEnabled}
>
```

gwwar: As an aside, how do we ever get into a disabled state?

src/boot/stylesheets/main.scss
kwight: It is needed, yes, since without it, it inherits the `cursor: pointer` property of `.wpnc__main button` (which seems a sensible default, being the most common state).

src/templates/button-back.jsx
kwight: > React is also smart enough to handle html5 boolean props correctly

Oh, cool – OK, since there's at least one other unrelated instance (to this PR) of this, I'll address it all together in a separate PR.

> As an aside, how do we ever get into a disabled state?

In only one situation: the Up/Down arrows on single view when at the top or bottom of the list.

src/boot/stylesheets/main.scss
gwwar: How about `cursor: default`? Using`pointer-events:none` is rather unusual and has some quirks in IE11.

kwight: That... makes so much more sense 🙃  Thanks!

PR #149, opened 2017-06-08
Fix: pre-commit hook wasn't prettifying files

### LINKED ISSUE #147: Prettier does not run automatically on commit.
Prettier doesn't run on commit; even with `jsx` files that have been changed, the `git diff` check [here](https://github.com/Automattic/notifications-panel/blob/master/bin/pre-commit#L7) always seems to result in no files (an empty `$files` variable).

Running `./bin/prettify-file.sh` manually identifies the changed files properly and corrects them.

dmsnell:
Resolves #147

It was discovered that no files were being prettified because the line
grabbing changed files always returned an empty result set.

The problem stemmed from the syntax of the Regular Expression in `grep`
and has been resolved by updating and adding the `-E` flag to accept
alternation groups.

Thanks @kwight for your sleuthing on this!

cc: @Automattic/lannister


kwight:
Works for me! 🚢

PR #150, opened 2017-06-08
Fix filter missing border/pixel shift when running in iframe #131

### LINKED ISSUE #131: Pixel jump on filter buttons
In the iframe (frontend), the filter buttons jump when you open the detail view (by clicking on any notification).

Safari Version 10.1.1 
Mac OS 10.12.5

note bug

roundhill:
It's a bit tricky using media queries within `notifications-panel`, because when we run in the iframe, the document's width is different because it's limited to the width of the iframe.

This PR removes the media query we were using to show/hide the border to the left of the filter bar. It now shows properly in wide mode in both the iframe and calypso:

_iframe_
iframe

_calypso_
calypso

There is one drawback, when on the small screen view you will have the border to the left. But to my eyes it is so small and hardly noticeable so I think it's a worthy trade off:
screen shot 2017-06-08 at 4 19 39 pm

**To test**
Run the client in the iframe and calypso. There should be a border to the left of the filter bar. When opening a note, the border should change to white.

Fixes #131

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -165,22 +165,17 @@ body {
     	background-color: $white;
     	color: $gray-text-min;
     	border-bottom: 1px solid lighten( $gray, 30 );
+        border-left: 1px solid $white;
     	text-align: center; // Center filter in IE 9
     	height: $filter-height;
     	box-sizing: border-box;
     	direction: ltr;
     	display: table; //fallback for browsers not supporting flexbox
-
-		@media only screen and (min-width: 661px) {
-			border-left: 1px solid $white;
-		}
     }
 
-	@media only screen and (min-width: 661px) {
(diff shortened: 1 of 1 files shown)


src/boot/stylesheets/main.scss
```diff
@@ -165,22 +165,17 @@ body {
     	background-color: $white;
     	color: $gray-text-min;
     	border-bottom: 1px solid lighten( $gray, 30 );
+        border-left: 1px solid $white;
```
dmsnell: extra space?

dmsnell:
pixel perfect!

src/boot/stylesheets/main.scss
roundhill: I think all of `main.scss` needs a reformat... or my editor config is messed up?

roundhill: Fixed it... kind of :) It matches the other lines around it at least.

PR #151, opened 2017-06-09
Run `main.scss` through Prettier with 4-spaced tabs.

kwight:


dmsnell:
Do we have separate settings for style files? I'd rather keep it all the same and just open up the pre-commit file to run over .scss files.

kwight:
Being handled in #152 .

PR #152, opened 2017-06-09
Prettify our stylesheets.

kwight:
Update Prettier to v1.4.0, to include prettifying our Sass stylesheets.

This keeps the existing `prettier` options the same; the affect on something like `main.scss` is everything being a space, with 4-space indents:

screen shot 2017-06-14 at 1 30 19 pm

Tested with `main.scss` and `status-bar.jsx`, no apparent regressions.


kwight:
Superceded by #159 .

PR #158, opened 2017-06-13
Fix responding to small screens when in iframe

roundhill:
There's a regression in the iframe version of `notificationsbeta` where notifications will not become full width on small screens. The fix was to not have a set width for the notes list until we reach a certain screen width.

Note: The spacing in `main.scss` hasn't been fixed up yet, so sorry for any weird spacing issues. See #152 

**To test**
* Run this branch on your sandbox: `./build-sandbox.sh -b -c fix/iframe-responsiveness`
* Open notifications on the front end of a sandboxed site
* Resize the browser window it should become full width at 480px or smaller.
* Also test the branch in calypso to ensure there aren't any regressions

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -796,11 +796,11 @@ body {
 
 		.wpnc__note-list {
 			left: auto;
-			width: 400px;
+			width: 410px;
 		}
 
     	.wpnc__single-view {
-    		right: 400px;
+    		right: 410px;
     		left: 10px;
     		top: 0px;
     		bottom: 0px;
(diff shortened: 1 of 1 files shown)


src/boot/stylesheets/main.scss
```diff
@@ -327,17 +327,12 @@ body {
       flex-direction: column;
       top: 0;
     	right: 0;
-    	left: auto;
+    	left: 0;
```
gwwar: Minor, but perhaps we can align these with `top` and `bottom`

kwight:
Ran through testing instructions, looked at CSS, everything looks good to go for me 👍

roundhill:
Thanks @gwwar and @kwight!

PR #159, opened 2017-06-14
Commit to two spaces for indenting.

kwight:
Let's do it – let's commit the project to the Prettier default of two spaces for indenting (this PR relies on and includes #152). Let the reign of confusion be over.

Examples of `main.scss` and `status-bar.jsx`:

screen shot 2017-06-14 at 1 36 00 pm

screen shot 2017-06-14 at 1 37 39 pm

**Testing**
* Using this PR, run `./bin/prettify-file.sh "src/boot/stylesheets/main.scss"` (or any other stylesheet).
* Run `./bin/prettify-file.sh "src/templates/status-bar.jsx"` (or any other JSX file).
* Check that there are no visual regressions, and that indenting is done with two true spaces.

### DIFF
--- a/bin/prettify-file.sh
+++ b/bin/prettify-file.sh
@@ -12,4 +12,4 @@ if [ ! -f ./node_modules/.bin/prettier ]; then
 	exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


kwight:
> is there a standard prettier config file we could store the options in?

Oh, I'm not sure; I'll look into that next.

Issue #154, opened 2017-06-09
React.createClass is deprecated and will be removed in React v16.


kwight:
> Warning: Localized(BackButton): React.createClass is deprecated and will be removed in version 16. Use plain JavaScript classes instead. If you're not yet ready to migrate, create-react-class is available on npm as a drop-in replacement.

Instances to be updated:

 - [x] ActionButton
 - [x] ActionsPane
 - [x] BlockComment
 - [x] BlockPost
 - [ ] BlockUser
 - [x] ButtonApprove
 - [x] ButtonLike
 - [x] ButtonSpam
 - [x] ButtonTrash
 - [ ] CommentReplyInput
 - [x] ContainerHotkey
 - [x] EmptyMessage
 - [x] Error
 - [x] FilterBar
 - [ ] FollowLink
 - [ ] Gridicons
 - [x] ImageLoader
 - [ ] Layout
 - [x] NavButton
 - [ ] NoteBody
 - [ ] NoteList
 - [x] Preface
 - [ ] StatusBar
 - [ ] Suggestion
 - [ ] SummaryInList
 - [ ] SummaryInSingle
 - [ ] UndoListItem
 - [ ] UserHeader

abhishek-2102:
Read this [#8854](https://github.com/facebook/react/issues/8854) 
You may get some idea about the warning.

kwight:
Closed in the above PRs. There is still some usage of `create-react-class` though.

PR #166, opened 2017-06-22
Update syntax in React classes to eliminate `React.createClass()`

dmsnell:
Resolves part of #154 

Should only be styling updates

### DIFF
--- a/src/templates/block-comment.jsx
+++ b/src/templates/block-comment.jsx
@@ -1,25 +1,17 @@
 import React from 'react';
+import classNames from 'classnames';
 
 import { html } from '../indices-to-html';
 import { p } from './functions';
 
-export const CommentBlock = React.createClass({
-  render: function() {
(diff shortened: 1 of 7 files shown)


src/templates/block-comment.jsx
```diff
@@ -1,25 +1,17 @@
...
-    }
-
-    return (
-      <div className={className}>
-        {commentText}
-      </div>
-    );
-  },
-});
+export const CommentBlock = ({ block, meta }) =>
```
kwight: The "export" here is extraneous.

dmsnell: this is how I write my modules. I make sure that every export is a named export, then I often provide a default export. this is partially to support a pattern and partially to support testing

kwight: AH! Carry on then :)

PR #167, opened 2017-07-05
Fix: Notes not loading

### LINKED ISSUE #162: Notifications sometimes don't appear when user is proxied.
We have some reports of proxied users not having their notifications displayed upon opening the panel. Refreshing, visiting another site and coming back, or reconnecting the proxy seem to "fix" it.

dmsnell:
Resolves #162 

It appears that when loading the notifications panel in a window that is originally `hidden` then the notes core never realizes that it has become visible. Thus, when we open it it's still not polling for the notifications data.

Now we are passing along current visibility information in the standalone client when changing window visibility. The `rest-client` then updates its own state based on this value.

**Testing**
Load the notifications panel when hidden. Reload a browser tab and switch to another tab, for instance. In **master** it should not load notes but in this branch it should when you switch back to the tab.

### DIFF
--- a/src/Notifications.jsx
+++ b/src/Notifications.jsx
@@ -61,7 +61,19 @@ export class Notifications extends PureComponent {
     initStore({
       customEnhancer,
       customMiddleware: mergeHandlers(customMiddleware, {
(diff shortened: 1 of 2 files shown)


aidvu:
Woohoo, good job!

PR #168, opened 2017-07-07
React: refactor usage of React.createClass()

dmsnell:
More work in #154 

**Testing**

As usual, just test with `npm start`. Any breakage should be obvious via errors or malformed UI.
There are no functional changes in this PR: only syntax.

Pay attention to the filter bar where I have actually refactored some logic to remove the `var` reassignment.

cc: @Automattic/lannister

### DIFF
--- a/src/templates/container-hotkey.jsx
+++ b/src/templates/container-hotkey.jsx
@@ -1,40 +1,40 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { Component, PropTypes } from 'react';
 
 /**
  * Internal dependencies
  */
 import { modifierKeyIsActive, shortcutsAreEnabled } from '../helpers/input';
 
-const HotkeyContainer = React.createClass({
-  propTypes: {
-    shortcuts: React.PropTypes.arrayOf(
-      React.PropTypes.shape({
-        action: React.PropTypes.func.isRequired,
-        hotkey: React.PropTypes.number.isRequired,
-        withModifiers: React.PropTypes.bool,
+const dispatch = (event, action) => {
+  event.preventDefault();
+  event.stopPropagation();
+
+  action();
+};
+
+export class HotkeyContainer extends Component {
+  static propTypes = {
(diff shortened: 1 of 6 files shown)


src/templates/empty-message.jsx
```diff
@@ -1,64 +1,53 @@
...
+    const { emptyMessage, link, linkMessage } = this.props;
 
     return (
       <div
         className="wpnc__empty-notes-container"
         style={{ height: this.props.height - TITLE_OFFSET + 'px' }}
       >
-        {message}
+        {link &&
+          linkMessage &&
```
kwight: Any reason you wouldn't keep these on the line above?

dmsnell: `prettier` did it

src/templates/container-hotkey.jsx
```diff
@@ -1,53 +1,53 @@
 /**
  * External dependencies
  */
-import React from 'react';
+import React, { Component, PropTypes } from 'react';
```
dmsnell: `React` is the default export while `Component` and `PropTypes` are named exports. They have different semantics than `require()`, so it's not right to `import React` and then use `React.Component`.

Copons: Also `React.PropTypes` is [deprecated](https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html#migrating-from-react.proptypes) as of React 15.5 where `PropTypes` will be moved in its own separate package:
`import PropTypes from 'prop-types';`

dmsnell: Yeah thanks @copons. I think it will be easy enough to automate that change later on. Should I be updating it now? I don't know.

Copons: There's a codemod ready for this: https://github.com/reactjs/react-codemod#react-proptypes-to-prop-types
Maybe it makes sense to just do it whenever we update to 15.5.

PR #171, opened 2017-07-25
Support for OPEN_COMMENT middleware action

samouri:
In order to support opening directly to comments within Calypso, the notifications-panel needs to support dispatching a new action for opening comments instead of just one for opening posts.

On top of that, we also need to start dispatching actions when timestamps are clicked.

This PR seeks to address both points.

### DIFF
--- a/src/templates/functions.jsx
+++ b/src/templates/functions.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import { html as toHtml } from '../indices-to-html';
+import { pickBy } from 'lodash';
 
 /**
  * Adds markup to some common text patterns
@@ -230,3 +231,27 @@ export function zipWithSignature(blocks, note) {
 }
 
 export const validURL = /^(?:http(?:s?)\:\/\/|~\/|\/)?(?:\w+:\w+@)?((?:(?:[-\w\d{1-3}]+\.)+(?:com|org|net|gov|mil|biz|info|mobi|name|aero|jobs|edu|co\.uk|ac\.uk|it|fr|tv|museum|asia|local|travel|blog|[a-z]{2}))|((\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)(\.(\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)){3}))(?::[\d]{1,5})?(?:(?:(?:\/(?:[-\w~!$+|.,=]|%[a-f\d]{2})+)+|\/)+|\?|#)?(?:(?:\?(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)(?:&(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)*)*(?:#(?:[-\w~!$ |\/.,*:;=]|%[a-f\d]{2})*)?$/i;
+
+export const linkProps = note => {
+  const { site, comment, post } = note.meta.ids;
+  let type;
+
+  if ( comment ) {
+    type = 'comment';
+  } else if ( post ) {
+    type = 'post';
+  }
+
+  switch (type) {
+    case 'post':
+    case 'comment':
+      return pickBy( {
+        'data-link-type': type,
+        'data-site-id': site,
+        'data-post-id': post,
+        'data-comment-id': comment,
+      } );
+    default:
+      return {};
+  }
+};
--- a/src/templates/functions.jsx
+++ b/src/templates/functions.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { html as toHtml } from '../indices-to-html';
-import { pickBy } from 'lodash';
+import { pickBy, get } from 'lodash';
 
 /**
  * Adds markup to some common text patterns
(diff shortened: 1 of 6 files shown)


gwwar:
cc @Automattic/lannister

samouri:
This seemed to work when i tested it out with https://github.com/Automattic/wp-calypso/pull/14396

src/templates/functions.jsx
```diff
@@ -230,3 +231,27 @@ export function zipWithSignature(blocks, note) {
 }
 
 export const validURL = /^(?:http(?:s?)\:\/\/|~\/|\/)?(?:\w+:\w+@)?((?:(?:[-\w\d{1-3}]+\.)+(?:com|org|net|gov|mil|biz|info|mobi|name|aero|jobs|edu|co\.uk|ac\.uk|it|fr|tv|museum|asia|local|travel|blog|[a-z]{2}))|((\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)(\.(\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)){3}))(?::[\d]{1,5})?(?:(?:(?:\/(?:[-\w~!$+|.,=]|%[a-f\d]{2})+)+|\/)+|\?|#)?(?:(?:\?(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)(?:&(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)*)*(?:#(?:[-\w~!$ |\/.,*:;=]|%[a-f\d]{2})*)?$/i;
+
+export const linkProps = note => {
+  const { site, comment, post } = note.meta.ids;
```
dmsnell: should probably make a safe get here because I'm not sure all notes are going to have these.

dmsnell:
Let's be careful with the `getPostRef()` function. I can't remember why I created it that way but you can definitely see that I'm checking in the note header _and_ in the note body. We need to make sure that we aren't getting notes with the comment or post links which don't have the meta.

src/templates/functions.jsx
samouri: only need a safe get if `meta.ids` isn't guaranteed.  I'm okay all of the 3 being undefined.

dmsnell: I was talking about `note.meta.ids`

gwwar:
```jsx
const { site, comment, post } = get( note, 'meta.ids', {} );
```

gwwar:
Nice @samouri I can get this working most of the time, but sometimes I see this glitch. Can you see if you can repro?

screen shot 2017-07-26 at 5 13 24 pm

src/templates/functions.jsx
samouri: makes sense. thanks for the example gwwar!

samouri:
@gwwar: I used to see that issue in two edgecases:

1. if the comment doesn't exist or for whatever reason the network requests 404 (permissions etc).
2. if the comment does exist but for a different post on the site

I believe i've fixed those, and haven't run into any more since.  I'll play with it more to see if i can find a repro

src/templates/functions.jsx
```diff
@@ -233,7 +233,7 @@ export function zipWithSignature(blocks, note) {
 export const validURL = /^(?:http(?:s?)\:\/\/|~\/|\/)?(?:\w+:\w+@)?((?:(?:[-\w\d{1-3}]+\.)+(?:com|org|net|gov|mil|biz|info|mobi|name|aero|jobs|edu|co\.uk|ac\.uk|it|fr|tv|museum|asia|local|travel|blog|[a-z]{2}))|((\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)(\.(\b25[0-5]\b|\b[2][0-4][0-9]\b|\b[0-1]?[0-9]?[0-9]\b)){3}))(?::[\d]{1,5})?(?:(?:(?:\/(?:[-\w~!$+|.,=]|%[a-f\d]{2})+)+|\/)+|\?|#)?(?:(?:\?(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)(?:&(?:[-\w~!$+|.,*:]|%[a-f\d{2}])+=?(?:[-\w~!$+|.,*:=]|%[a-f\d]{2})*)*)*(?:#(?:[-\w~!$ |\/.,*:;=]|%[a-f\d]{2})*)?$/i;
 
 export const linkProps = note => {
-  const { site, comment, post } = note.meta.ids;
+  const { site, comment, post } = get( note, 'meta.ids', {} );
```
gwwar: Don't forget to import get from lodash

samouri: 🧀 thanks

gwwar:
Hmm, I wasn't able to repro this morning. We can write up an issue if other folks run into it. Once the import is fixed, let's 🚢 

Nice work @samouri

gwwar:
Oh and, let's also bump the package version 😄

samouri:
@gwwar: package version bumped to `1.2.0` b/c this change is breaking.
🚢 ?

samouri:
> Let's be careful with the getPostRef() function. I can't remember why I created it that way but you can definitely see that I'm checking in the note header and in the note body. We need to make sure that we aren't getting notes with the comment or post links which don't have the meta.

I'll be on high alert for bugs for the next week - and if anything happens we know what to check!

PR #172, opened 2017-07-27
Links: open in new tab if meta/ctrl are pressed

samouri:
~~For now this should have no functional changes.~~

~~It adds the property `newTab` to the OPEN_POST / OPEN_LINK actions.  newTab true/false is determined by if ctrl/meta are on the event object.~~

On the suggestion of @dmsnell, this ignores sending anything to the middleware and opens links in a new tab when cmd/ctrl are held during the notification click.

mdn source for mouse events: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

To test:
In the calypso directory run: `npm install 'Automattic/notifications-panel#fix/cmdclick-open-new-tab'`
then make run as usual and verify cmd/ctrl clicks are working as expected

### DIFF
--- a/src/utils/link-interceptor.js
+++ b/src/utils/link-interceptor.js
@@ -25,6 +25,12 @@ export const interceptLinks = event => {
     return true;
   }
 
+  const newTab = event.ctrlKey || event.metaKey;
+  if (newTab) {
+    window.open(href, '_blank');
+    return true;
+  }
+
   event.stopPropagation();
   event.preventDefault();
 
--- a/src/utils/link-interceptor.js
+++ b/src/utils/link-interceptor.js
@@ -27,8 +27,7 @@ export const interceptLinks = event => {
 
   const newTab = event.ctrlKey || event.metaKey;
   if (newTab) {
-    window.open(href, '_blank');
-    return true;
+    return true; // will open in new tab b/c we dont eat the event
   }
 
   event.stopPropagation();
--- a/src/utils/link-interceptor.js
+++ b/src/utils/link-interceptor.js
@@ -25,9 +25,12 @@ export const interceptLinks = event => {
     return true;
   }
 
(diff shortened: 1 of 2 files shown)


samouri:
some individuals have brought up the possibility of ignoring the middleware and in case of `newTab`, don't send through the action and just launch a window.open

src/utils/link-interceptor.js
```diff
@@ -25,6 +25,11 @@ export const interceptLinks = event => {
     return true;
   }
 
+  const newTab = event.ctrlKey || event.metaKey;
+  if (newTab) {
+    return true; // will open in new tab b/c we dont eat the event
```
samouri: question: does returning true/false matter here? Its just an onClick handler for a div and I can't tell that anything responds to it...

dmsnell: if it's an `<a>` it will cause the browser to respond differently.

probably not on a `<div>` but I do like to keep the pattern intact.

gwwar:
Yeah probably 1.2.1 over 1.3.0 here 👍

samouri:
switched the package version over to 1.2.1 instead of 1.3

src/utils/link-interceptor.js
dmsnell: @samouri I think this comment and the naming could be a bit clearer here, especially since the link interceptor itself doesn't know about new tabs…

if we can rename and update the comments I think the PR will be ready. here's an idea

```js
// we don't want to interfere with the click
// if the user has specifically overwritten the
// normal behavior already by holding down
// one of the modifier keys.
if ( event.ctrlKey || event.metaKey ) {
	return true;
}
```

dmsnell:
@samouri all good. :shipit: I haven't tested these changes but will attempt to do so before we update the iframe or Calypso version.

PR #174, opened 2017-08-03
package.json: Change main to esnext

ockham:
To better denote we're dealing with untranspiled code.

Companion Calypso PR: https://github.com/Automattic/wp-calypso/pull/16870

To provide a bit of an explanation -- `esnext` is a custom field.
* We'll specify it as a possible entry point for webpack to watch out for [here](https://github.com/Automattic/wp-calypso/pull/16870/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR141).
* OTOH, we use it (as opposed to `main`) as a flag to indicate that this is a package's entrypoint for untranspiled code [here](https://github.com/Automattic/wp-calypso/pull/16870/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR108).

This is a strategy found at http://2ality.com/2017/06/pkg-esnext.html.

Suggested order:
* Merge this
* Release new version
* Calypso: Include `notifications-panel` version bump with https://github.com/Automattic/wp-calypso/pull/16870
* Test https://github.com/Automattic/wp-calypso/pull/16870
* Merge https://github.com/Automattic/wp-calypso/pull/16870

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "2.0.0",
+  "version": "1.3.0",
   "description": "The core notifications panel for WordPress.com notifications",
   "esnext": "src/Notifications.jsx",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "notifications-panel",
   "version": "1.2.7",
   "description": "The core notifications panel for WordPress.com notifications",
-  "main": "src/Notifications.jsx",
+  "esnext": "src/Notifications.jsx",
   "scripts": {
     "build": "webpack",
     "build:prod": "NODE_ENV=production webpack -p",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.2.7",
+  "version": "2.0.0",
   "description": "The core notifications panel for WordPress.com notifications",
   "esnext": "src/Notifications.jsx",
   "scripts": {


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "notifications-panel",
   "version": "1.2.1",
```
dmsnell: this is probably worth a major version bump, no? since software already using this would have to update in order not to break?

dmsnell:
@gwwar @kwight I have updated this with a major version bump (see the commit description).

Could you two consider this change and indicate if you have any hesitation or questions? This is a change to help @ockham improve Calypso's build process by automatically inferring which `node_modules` can be included as-is without having to import the already-transpiled builds (which would end up speeding up our build _and_ saving on build-size).

kwight:
@ockham do you have a strict timeline for this? We've moved to scheduled releases for `notifications-panel` (p3fqKv-5V6-p2), so this could be part of the next beta on Nov 8th, with production release Nov 22nd; would that work for you?

gwwar:
To add to that Calypso gets the beta release so unless it's really urgent, it'd be slated for Nov 8.

ockham:
No strict/urgent timeline, but would totally appreciate this being merged!

gwwar:
Thanks @ockham we'll try to test/merge this soon, though for now we'll stick with the Nov 8 version bump.

kwight:
I've committed these changes both to `master` (by merging this PR with a minor bump to 1.3.0) and to the `react16` branch (with a bump to 2.1.0) in 011fc14a6da274eb238eda295f2b40003d25ff73.

ockham:
D'oh, this won't work until we merge https://github.com/Automattic/wp-calypso/pull/19698 

Quoting https://github.com/Automattic/wp-calypso/pull/19723#issuecomment-347925487:

> To fix, we can:
> * either merge #19698 
> * or change back `esnext` to `module` in https://github.com/Automattic/notifications-panel
> * or add a build step, publish that to npm, and set  `main` to point to the dir that holds the transpiled code

PR #175, opened 2017-08-03
i18n: Fix display of English text in the filter bar

### LINKED ISSUE #173: i18n: Top navigation bar loads in English, switches to correct language after first click
I have my UI language set to German, yet the header loads in English.

Here is a demo:
notifications-translation

akirk:
Fixes #173. The component wasn't started localized.

Old behavior:
notifications-translation

New behavior: No change of UI language.


akirk:
@dmsnell thanks for the review! How will this get deployed?

PR #177, opened 2017-08-03
Middleware/Links: add OPEN_SITE

samouri:
We already send out OPEN_COMMENT and OPEN_POST.  It only naturally follows that we should also have OPEN_SITE.  

Question: are there any others that are missing that should open in Reader?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.2.1",
+  "version": "1.3.0",
   "description": "The core notifications panel for WordPress.com notifications",
   "main": "src/Notifications.jsx",
   "scripts": {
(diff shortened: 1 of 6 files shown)


samouri:
To Test in Calypso:
0. go to master branch and make sure you have latest
1. npm install 'Automattic/notifications-panel#add/open-site-action'
2. make start

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.2.1",
+  "version": "1.3.0",
```
samouri: it changes the api. 

hypothetically if a client were expecting OPEN_LINKs they will be unpleasantly surprised that now they are getting OPEN_SITE

dmsnell: yeah, make your call. no client currently would be surprised I think

dmsnell: to be consistent though, we just made this decision in a recent PR along these "wrong" rules

PR #179, opened 2017-08-16
Build: Use long `git` description

dmsnell:
When a commit hangs exactly at a tag `git describe --always --dirty`
will return just the tag and not the commit hash. This update makes sure
that we always include a commit hash in the output.

**Testing**

Just build it and look at `index.html` - I did and it looks right.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "1.2.2",
+  "version": "1.2.3",
   "description": "The core notifications panel for WordPress.com notifications",
   "main": "src/Notifications.jsx",
   "scripts": {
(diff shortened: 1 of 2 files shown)


samouri:
👀 checking this out and also looking up what the heck a dirty long description means

gwwar:
@samouri https://git-scm.com/docs/git-describe

--dirty
> Describe the state of the working tree. When the working tree matches HEAD, the output is the same as "git describe HEAD". If the working tree has local modification "-dirty" is appended to it...

--long
> Always output the long format (the tag, the number of commits and the abbreviated commit name) even when it matches a tag. This is useful when you want to see parts of the commit object name in "describe" output, even when the commit in question happens to be a tagged version. Instead of just emitting the tag name, it will describe such a commit as v1.2-0-gdeadbee (0th commit since tag v1.2 that points at object deadbee…​.).

PR #184, opened 2017-10-11
Modify exports to make webpack es6 happy

samouri:
for use in https://github.com/Automattic/wp-calypso/pull/16057 and potentially an eventual merge

### DIFF
--- a/src/comment-replies-cache/index.js
+++ b/src/comment-replies-cache/index.js
@@ -80,13 +80,16 @@ function cleanupRepliesCache() {
   });
 }
 
-var LocalStorageMixin = {
+export const LocalStorageMixin = {
   localStorage: {
     getItem: getItem,
     setItem: setItem,
     removeItem: removeItem,
   },
 };
+export const cleanup = cleanupRepliesCache;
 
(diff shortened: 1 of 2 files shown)


src/comment-replies-cache/index.js
```diff
@@ -80,13 +80,16 @@ function cleanupRepliesCache() {
...
   },
 };
+export const cleanup = cleanupRepliesCache;
 
-module.exports.cleanup = cleanupRepliesCache;
-module.exports.LocalStorageMixin = LocalStorageMixin;
+export default {
+  cleanup,
+  LocalStorageMixin,
+};
```
psealock: Whats the reason to export them individually and as a `default`ed object?

samouri: Its so that I don't need to chase down how the consumers are using it. Definitely not something that should be done going forward

samouri:
>  Looks good to me if you tested it out. I didn't.


I didn't do an extensive tests but I also don't think this change calls for it.
Webpack built successfully and basic notifications-panel functionality seemed to work 👍 

I'll merge

PR #187, opened 2017-10-18
Gridicon: updates the reply icon

### LINKED ISSUE #186: Reply icon should be pointing to the right
Following this change: https://github.com/Automattic/gridicons/pull/255, the Reply gridicon should now be pointing to the right like so:

screenshot 2017-10-18 09 59 38

They currently look like this in these pages:

1. My Sites > Comments

screenshot 2017-10-18 10 28 26

2. Reader
screenshot 2017-10-18 10 25 11

But points to the left in Notifications:
screenshot 2017-10-18 10 29 53

rodrigoi:
Fixes #186 

Updates the reply icon for consistency with the latest release of gridicons, since we are not using the npm package, but a subset of icons on `src/templates/gridicons.jsx`

https://github.com/Automattic/notifications-panel/blob/5ef66103dfcbabed9de0353d9ea27bb718cd6386/src/templates/gridicons.jsx#L192

Maybe we should consider using the npm package instead...

To test:
- run the local environment
- find a notification of a comment with a reply
- both the list view and the detail view should show the reply icon pointing to the right.

| Before | After |
|--|--|
| screen shot 2017-10-18 at 15 14 57 | screen shot 2017-10-18 at 15 12 43 |
| screen shot 2017-10-18 at 15 15 23 | screen shot 2017-10-18 at 15 12 54 |

References:
https://github.com/Automattic/gridicons/pull/255
https://github.com/Automattic/wp-calypso/pull/18891

### DIFF
--- a/src/templates/gridicons.jsx
+++ b/src/templates/gridicons.jsx
@@ -194,7 +194,7 @@ export default React.createClass({
           <svg {...sharedProps} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
             <title>Reply</title>
             <g>
-              <path d="M14 8H6.828l2.586-2.586L8 4 3 9l5 5 1.414-1.414L6.828 10H14c2.206 0 4 1.794 4 4s-1.794 4-4 4h-2v2h2c3.314 0 6-2.686 6-6s-2.686-6-6-6z" />
+              <path d="M9 16h7.2l-2.6 2.6L15 20l5-5-5-5-1.4 1.4 2.6 2.6H9c-2.2 0-4-1.8-4-4s1.8-4 4-4h2V4H9c-3.3 0-6 2.7-6 6s2.7 6 6 6z" />
             </g>
           </svg>
         );


kwight:
> Maybe we should consider using the npm package instead...

If we're only using a handful (looks like roughly 1/8 of them), and they change fairly rarely, I'd vote for us continuing to save the bits transferred and keep manually patching until it becomes more of an issue.

dmsnell:
Last time I checked, the gridicons were written in such a way that they were all bundled up together and the `npm` import was disproportionately huge compared to our needs.

On the other hand, if it's updated, or if we update it, so that it can also create ES modules and we can import only the ones we want and let webpack trim down the bundle then that's a more stable long-term solution

jancavan:
I'm still seeing the old icon?

screenshot 2017-10-19 09 44 03

rodrigoi:
> I'm still seeing the old icon?

this hasn't been deployed yet. Will let you know when it is 😉

kwight:
@jancavan This is in production now 👍

jancavan:
@kwight @rodrigoi Looks good, thanks!

PR #189, opened 2017-10-21
Add cart icon to the list of allowed noticon2gridicons

justinshreve:
This PR adds the `cart` icon to the noticon to gridicon transformation list, making it possible to use the `cart` icon for notifications.

We will be using this for a [new order notification for Store on WP.com](https://github.com/Automattic/wp-calypso/issues/16181).

screen shot 2017-10-21 at 12 24 33 pm

To Test:
* Install `notifications-panel` per the [README](https://github.com/Automattic/notifications-panel#installation).
* Load up this branch.
* Sandbox `public-api`.
* Open `class-notification-builders-v1-1.php`and find the builder for one of the notifications in your stream. For example, `build_post_reply` for a comment reply. 
* Update the `set_noticon` call to `$builder->set_noticon( 'cart' );`
* Load up the notifications panel (`http://notifications.dev:8888/`) and view.
* Verify that you can see the cart icon.
* (You could also load up D7802-code and generate an actual order notification - but I haven't completely written out the testing steps there and they are a bit more involved).

### DIFF
--- a/src/templates/gridicons.jsx
+++ b/src/templates/gridicons.jsx
@@ -252,6 +252,17 @@ export default React.createClass({
             </g>
           </svg>
         );
+
+      case 'gridicons-cart':
+        return (
+          <svg {...sharedProps} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+            <title>Cart</title>
+            <g>
+              <path d="M9 20c0 1.1-.9 2-2 2s-1.99-.9-1.99-2S5.9 18 7 18s2 .9 2 2zm8-2c-1.1 0-1.99.9-1.99 2s.89 2 1.99 2 2-.9 2-2-.9-2-2-2zm.396-5c.937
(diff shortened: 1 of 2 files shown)


kwight:
Sweet, looks and works great 👍

justinshreve:
Thanks for the reviews!

@kwight Just wondering, what is the process for getting changes deployed? Does Lannister handle that? I'm not in a huge rush (I still have some work to do on the .com side) but I wasn't sure if there was anything I needed to do to push it live.

kwight:
@justinshreve We've just started implementing a two-week rotating schedule for releases (p3fqKv-5V6-p2); I can cut a new beta today, giving us two weeks until production release Nov8th. Of course, if that doesn't work, just let me know and we'll certainly work something out!

justinshreve:
@kwight That should work perfectly I think. Thank you!

PR #192, opened 2017-11-02
Migrate PropType references to external package to solve console warnings

eliorivero:
This PR solves these warnings in console:
```
Warning: Accessing PropTypes via the main React package is deprecated, and will be removed in  React v16.0. Use the latest available v15.* prop-types package from npm instead.
```


kwight:
Thanks! We've got this covered though in #191 , our `react16` branch (tagged to v2.0.0).

PR #197, opened 2017-11-15
Add support for checkmark icon

dmsnell:
Despite having the `checkmark` Gridicon checked-in to the project the
mapping from noticon index was missing in `noticon2gridicon`.

This patch adds the mapping and updates the patch version. No changes to
the code or operating are affected by this: it's only a bug-fix.

**Testing**

Find a note with the checkmark icon in it and see if it displays as a
checkmark or as an info `i`

**Before**
screen shot 2017-11-15 at 5 26 53 pm

**After**
screen shot 2017-11-15 at 5 26 30 pm

**Creating a checkmark note**

Open up any note building in your sandbox and change the icon type to `checkmark`

Also, Jetpack monitor and auto-update notices use the checkmark, so it could be
possible to create one by taking down a Jetpack site temporarily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,3 +6,4 @@
 
  - Center images in notification render. This was already happening for badge notifications and for
    all images in the mobile apps, so this change is somewhat a harmonization more than a change.
+ - Support `checkmark` icon type in notes list.
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -847,7 +847,11 @@ body {
         margin-top: 0px;
       }
 
-      -webkit-transform: translate3d(0, 0, 0); // fix for getting scrollbar in right z-index
+      -webkit-transform: translate3d(
+        0,
+        0,
+        0
+      ); // fix for getting scrollbar in right z-index
     }
 
     .wpnc__single-view {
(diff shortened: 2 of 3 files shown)


dmsnell:
This should coordinate with #196 if we add the changeling; also we can probably combine this and that one into one version bump if we want; they are both very small changes I think.

gwwar:
Changes here look fine, but could you add instructions on how to generate a note with a checkmark? @dmsnell we're also on a 2 week release schedule. So 2nd and 4th Wednesdays of the month. Did you need this sooner?

dmsnell:
> So 2nd and 4th Wednesdays of the month. Did you need this sooner?

Is that today? that's more up to @singerb and those of us with activity log work. I guess if it's not problematic to do an off-interval update I don't mind releasing it. if problematic we can wait.

gwwar:
For this month it'll be the 22nd. I'll leave it up to @kwight if he has extra cycles or not :)

dmsnell:
rebased this again #196 so I could add the content to the changelog.

@kwight is there anything that you normally do in a release that I could do for you here?

singerb:
I don't think it's super problematic to wait, especially since there are existing notifications already living with this. Since they continue to work, no code changes are needed on the notification creation end, and the info icon is a reasonable substitute, this can probably be released when appropriate.

kwight:
Since we don't have anything on beta right now, I'll cut these two PRs into a beta today or tomorrow, and if there are no issues, we can then use next week's 22nd release to move it to prod.

kwight:
Didn't get to it today, but will push out the beta tomorrow.

PR #202, opened 2017-12-06
bump react and remove esnext changes

gwwar:
We were maintaining a 2.x line for React 16, now that this is in Calypso, let's bump the React version to match, and merge the react16 feature branch as a follow up step.

This also reverts esnext changes to notifications-panel since https://github.com/Automattic/wp-calypso/pull/19698 has not landed yet and is blocking an update of notifications-panel in Calypso.

@ockham let us know when you're ready for another try.

### Testing Instructions
- `npm install`
- `npm start`
- Local testing environment loads as expected.


ockham:
Thanks!

PR #203, opened 2017-12-06
Merge the `react16` branch back into `master`.

kwight:
Now that both Calypso and this repo are running on React 16, it would be best to merge the temporary React 16-based branch back into `master`. Note that this brings together React 16 (now 16.2) with the `esnext` changes originally introduced in #174 .


ockham:
Can we merge this without the `esnext` change? I'm afraid that's still not quite ready on the Calypso side. I don't think we should block `notifications-panel`'s React 16 upgrade by that though...

gwwar:
Thanks for the heads up @ockham

kwight:
Sure, removed the `esnext` changes from #174 in a981725a11e78d474dc3abcd960233b5ba7f32db.

Issue #170, opened 2017-07-13
Mentions: Enter key doesn't work for active @mention


beaulebens:
When in the notifications panel, and replying to a comment, you can "@mention" someone else on that site. In other places (e.g. the Calypso Editor) you can start typing a name after the `@` and then hit `Enter` to select the highlighted user. In the Notifications panel, that doesn't seem to work, and I think I am just getting the first name in the full list (not the filtered list even).

Example; start replying to a comment:
<img width="398" alt="screen_shot_2017-07-13_at_5_03_40_pm" src="https://user-images.githubusercontent.com/108942/28191434-81746fd4-67ed-11e7-8985-3248410ff46d.png">

See that the person you want to mention is highlighted, and hit enter to select them, and get someone else entirely (I believe the first person in the dataset driving the auto-complete):
<img width="394" alt="screen_shot_2017-07-13_at_5_04_06_pm" src="https://user-images.githubusercontent.com/108942/28191458-9c5247cc-67ed-11e7-9bb8-668c539137cb.png">

Note that the inserted user's username does not match the search query at all.

If you use the mouse to click the name instead, then it works as expected.

I'm on Chrome/macOS.

kwight:
Fixed in #204 .

PR #204, opened 2017-12-08
Mentions: various bugfixes

rodrigoi:
This PR fixes these bugs on the mentions mixin:

* #170 enter key to select active mention.
* Adds a space after the username to improve writing flow.
* moves suggestion fetching to `componenDidMount` to remove delay after first mention.
* suggestions now position correctly depending on the available space on the screen.
* removes flash of suggestion list when repositioning.
* when using arrow shortcuts, the list reposition itself to show hidden elements correctly.
* some minor cleanups in the code.
* ~~bonus: `package-lock.json`.~~

How to test:

Just use add mentions on the reply field of any comment. The list should appear almost instantaneously, selection with both mouse and keyboard should work as expected and the list should position itself either above or below the reply field depending on the available space on the page.


kwight:
> suggestions now position correctly depending on the available space on the screen

I could occasionally end up with suggestions up high and obscured; I couldn't really see a pattern:

screen shot 2017-12-08 at 11 36 46 am

Also be sure to give a rebase on master, now that the React 16 branch stuff is in (I tested rebased, everything still looked fine).

Thanks for working on this <3

rodrigoi:
@kwight I can't reproduce that problem. Tried scrolling while typing and while moving with the arrow keys, resizing the window and I just couldn't break it. I've opened issue #205 so we can work on that later.

gwwar:
Should we commit the package-lock.json in a separate PR?

rodrigoi:
> Should we commit the package-lock.json in a separate PR?

we should. Created #206 just with the lock file.

src/suggestions/index.jsx
```diff
@@ -72,14 +74,16 @@ export const SuggestionsMixin = {
     window.addEventListener('keydown', this.handleSuggestionsKeyDown, false);
     window.addEventListener('keyup', this.handleSuggestionsKeyUp, false);
     window.addEventListener('blur', this.handleSuggestionBlur, true);
+
+    this.props.fetchSuggestions(this.props.note.meta.ids.site);
   },
 
   componentDidUpdate() {
-    if (!this.suggestionMixin_suggestionList) {
+    if (!this.suggestionsMixin_suggestionList) {
```
gwwar: We may just want to rename this later. Such a long and odd mix of casing here 😅

rodrigoi: yeah, I started moving this to a high order component because the `mixing` thing it's so `2016`, but the diff was going to be huge. When we move it to the `hoc`, this will be so much clearer.

src/suggestions/index.jsx
```diff
@@ -88,6 +92,10 @@ export const SuggestionsMixin = {
     const textArea = this.replyInput;
     const textAreaClientRect = textArea.getBoundingClientRect();
 
+    this.suggestionsAbove =
+      suggestionList.offsetHeight > window.innerHeight - textAreaClientRect.top &&
+      suggestionList.offsetHeight < textAreaClientRect.top;
```
gwwar: Is there any way to add a check before this? Just a note that this will trigger a reflow when we measure node dimensions.

src/suggestions/index.jsx
```diff
@@ -142,11 +146,11 @@ export const SuggestionsMixin = {
     const endString = this.state.value.slice(caretPosition);
 
     this.setState({
-      value: startString + suggestion.user_login + endString,
+      value: startString + suggestion.user_login + ' ' + endString,
       suggestionsVisible: false,
     });
 
-    this.setCaretPosition(element, startString.length + suggestion.user_login.length);
+    this.setCaretPosition(element, startString.length + suggestion.user_login.length + 1);
```
gwwar: Another const here?
```jsx
const updatedValue = `${ startString }${ $userLogin } ${ endString }`;
//...
this.setCaretPosition( element, updatedValue.length );
```

src/suggestions/index.jsx
```diff
@@ -210,15 +209,16 @@ export const SuggestionsMixin = {
...
+    const matcher = queryMatcher(query);
+    const suggestions = this.props.suggestions
+      .filter(({ name }) => matcher.test(name))
+      .slice(0, 10);
 
     this.setState({
       suggestionsQuery: query,
       suggestionsVisible: typeof query === 'string',
-      selectedSuggestionId: null,
+      selectedSuggestionId: suggestions.length > 0 ? suggestions[0].ID : null,
```
gwwar: Do we have lodash `get` here?

src/suggestions/index.jsx
```diff
@@ -235,70 +235,48 @@ export const SuggestionsMixin = {
...
+    if (!suggestionsVisible || !suggestions.length) {
+      return;
     }
 
-    const selectedSuggestionId = this.state.selectedSuggestionId || suggestions[0].ID;
-
     return (
       <div
         className="wpnc__suggestions"
+        ref={div => (this.suggestionsMixin_suggestionList = div)}
```
gwwar: Minor note that we try and avoid defining functions in render.

src/suggestions/index.jsx
```diff
@@ -24,6 +23,13 @@ const KEY_DOWN = 40;
  */
 const suggestionMatcher = /(?:^|\s)@([^\s]*)$/i;
 
+/**
+ * This pattern looks for a query
+ *
+ * @type {RegExp} matches @query
+ */
+const queryMatcher = query => new RegExp(`^${query}| ${query}`, 'i'); // start of string, or preceded by a space
```
dmsnell: not sure what `query` should be but this will treat it as a RegExp object. that is, if `query === '[abc]'` then it will match only a single-letter string of `a`, `b`, or `c`.

lodash has a useful function to escape strings we inject into a RegExp object: `lodash.escapeRegExp( string )` (no surprise in the name, lol)

might be useful to reflect on what our intent here is

src/suggestions/index.jsx
dmsnell: we probably want to move this into its own function (and the `onMouse…` events below too) to avoid creating new functions on every render. I could imagine some issues with this here.

fun fact! `div` in this case could also be `null`, which is our indication that it's unmounting

src/suggestions/index.jsx
```diff
@@ -235,70 +235,48 @@ export const SuggestionsMixin = {
...
           {suggestions.map(suggestion =>
             <Suggestion
-              ref={this.suggestionsMixin_storeSuggestionNode}
-              data-suggestion-id={suggestion.ID}
               key={'user-suggestion-' + suggestion.ID}
+              getElement={suggestionElement =>
+                (this.suggestionsMixin_suggestionNodes = {
+                  ...this.suggestionsMixin_suggestionNodes,
+                  [suggestion.ID]: suggestionElement,
+                })}
```
dmsnell: same note here - do we have a good reason to move this into the JSX vs adding another method to the mixin?

dmsnell:
> Do y'all think those are blockers?

Not at all! I think this PR is a net improvement and is large enough to be its own unit. Pulling in the other work is probably going to involve its own difficult questions for the standalone version.

PR #206, opened 2017-12-11
Notifications: adds package-lock.json file to the project

rodrigoi:
Just adds `package-lock.json` to the project.


gwwar:
Let's add an engines section in the package.json as well. Very loose is fine, but let's note that we expect a major version of Node 8

PR #207, opened 2018-01-03
Change the local development domain from `.dev` to `.localhost`.

kwight:
Up to now, we have been using `notifications.dev` as the domain for local development. However, the `.dev` TLD is now valid, and [Google has forced its forwarding to HTTPS](https://ma.ttias.be/chrome-force-dev-domains-https-via-preloaded-hsts/) as of Chrome 63. This patch instead has us use the `.localhost` TLD (same as Calypso).

I've chosen to add a new app specific to `.localhost` rather than updating the existing app, since this is a Chrome-only issue so far.

**Testing**

1. Load `master` at http://notifications.dev:3000 in Chrome 63+, and get a browser error after being redirected to the HTTPS-equivalent URL.
2. Switch to this branch, and load http://notifications.localhost:3000 . You should not be redirected, and instead see the prompt for OAuth. Continue, and notifications should load properly.

### DIFF
--- a/standalone/index.js
+++ b/standalone/index.js
@@ -21,44 +21,57 @@ const customMiddleware = {
   APP_RENDER_NOTES: [
     (store, { latestType, newNoteCount }) =>
       newNoteCount > 0
-        ? sendMessage({ action: 'render', num_new: newNoteCount, latest_type: latestType })
+        ? sendMessage({
+            action: 'render',
+            num_new: newNoteCount,
+            latest_type: latestType,
+          })
         : sendMessage({ action: 'renderAllSeen' }),
   ],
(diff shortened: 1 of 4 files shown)


standalone/index.js
```diff
@@ -21,34 +21,47 @@ const customMiddleware = {
...
+      }),
   ],
   VIEW_SETTINGS: [() => window.open('https://wordpress.com/me/notifications')],
 };
 
 const render = () => {
   ReactDOM.render(
     React.createElement(AuthWrapper(Notifications), {
-      clientId: 52716,
+      clientId: 56641,
```
dmsnell: nice

PR #209, opened 2018-01-05
Identify the current user for Tracks stats.

kwight:
Some Tracks stats are currently being recorded as anonymous in the Notifications iframe (this is not an issue for Calypso, since `notifications-panel` is used directly as a module). This PR allows Tracks to run `identifyUser` from `w.js`, properly adding the current user to any subsequent `t.gif` requests.

**Testing - this repo**
* Load the standalone app with this PR, and open devtools to the Network tab.
* Filter for `t.gif`, and click on a note to open it.
* The resulting `t.gif` request should include `_ui` and `_ul` params with your user ID and username respectively.

**Testing - dotcom iframe**
* Build a fresh beta `notifications-client` on your sandbox, using `bin/build-sandbox.sh -bc add/tracks-user` and the repo's own instructions.
* Sandbox `widgets.wp.com`.
* Load the front-end of your sandboxed test site.
* Note that a successful call to the `/me` endpoint was made.
* Click on a note to open it, and check that a `t.gif` was requested, containing the same params mentioned in the above repo testing instructions.

### DIFF
--- a/standalone/auth-wrapper.jsx
+++ b/standalone/auth-wrapper.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import wpcom from 'wpcom';
 import proxyRequest from 'wpcom-proxy-request';
 
-const getStoredToken = () => {
+export const getStoredToken = () => {
     try {
         const [expiry, token] = localStorage.getItem('auth').split(':', 2);
 
--- a/standalone/auth-wrapper.jsx
+++ b/standalone/auth-wrapper.jsx
@@ -85,10 +85,21 @@ export const AuthWrapper = Wrapped => class extends Component {
         window.location.replace(uri);
     };
 
+    setTracksUser = () =>
+        this.state.wpcom
+        .me()
+        .get( { fields: 'ID,username' } )
+        .then( ( { ID, username } ) => {
+            window._tkq = window._tkq || [];
+            window._tkq.push( [ 'identifyUser', ID, username ] ) ;
+        } )
+        .catch( () => {} );
+
     render() {
         const { clientId, redirectPath, ...childProps } = this.props;
 
         if (this.state.wpcom) {
+            this.setTracksUser();
             return <Wrapped {...childProps} {...{ wpcom: this.state.wpcom }} />;
         }
 
--- a/standalone/auth-wrapper.jsx
+++ b/standalone/auth-wrapper.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import wpcom from 'wpcom';
 import proxyRequest from 'wpcom-proxy-request';
 
-export const getStoredToken = () => {
+const getStoredToken = () => {
     try {
         const [expiry, token] = localStorage.getItem('auth').split(':', 2);
 
--- a/standalone/auth-wrapper.jsx
+++ b/standalone/auth-wrapper.jsx
@@ -66,6 +66,12 @@ export const AuthWrapper = Wrapped => class extends Component {
         });
     }
 
+    componentDidUpdate(prevProps, prevState) {
+      if (!prevState.wpcom && this.state.wpcom) {
+        this.setTracksUser();
+      }
+    }
+
     maybeRedirectToOAuthLogin = () => {
         if (this.state.oAuthToken) {
             return this.setState({ wpcom: wpcom(this.state.oAuthToken) });
@@ -99,7 +105,6 @@ export const AuthWrapper = Wrapped => class extends Component {
         const { clientId, redirectPath, ...childProps } = this.props;
 
         if (this.state.wpcom) {
-            this.setTracksUser();
             return <Wrapped {...childProps} {...{ wpcom: this.state.wpcom }} />;
         }
 
--- a/standalone/index.js
+++ b/standalone/index.js
@@ -1,10 +1,8 @@
 import ReactDOM from 'react-dom';
 import React from 'react';
-import wpcom from 'wpcom';
-const debug = require('debug')('notifications:note');
 
 import Notifications, { refreshNotes } from '../src/Notifications';
-import AuthWrapper, { getStoredToken } from './auth-wrapper';
+import AuthWrapper from './auth-wrapper';
 import { receiveMessage, sendMessage } from './messaging';
 
 require('../src/boot/stylesheets/style.scss');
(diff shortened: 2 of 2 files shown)


standalone/index.js
```diff
@@ -1,114 +1,132 @@
...
+            fields: "ID,username"
+        },
+        (error, data) => {
+            if (error) {
+                return;
+            }
+            window._tkq = window._tkq || [];
+            window._tkq.push(["identifyUser", data.ID, data.username]);
+        }
+    );
```
dmsnell: I'm mixed here. In general I approve of this PR, but I feel like we're just duplicating code needlessly by coupling it in to `rest-client` as well. if we have access to `wpcom` here (which I think in some way we might) I'd just as soon see us code it all here because it's not part-and-parcel of the panel itself.

also, I believe we can use the Promise interface with the `wpcom` object.

```js
wpcom()
	.req
	.get( { path: '/me', apiVersion: '1.1' }, { fields: 'ID,username' } )
	.then( ( { ID, username } ) => ( window._tkq = window._tkq || [], window._tkq.push( [ 'identifyUser', ID, username ] ) ) )
```

kwight: Ah, I see your point; agreed, I'll keep it all in `standalone/index.js`.

kwight:
Noting that from what I can tell testing so far, this does _not_ work when contained in an iframe, the use-case for which we started down this road to begin with. It's still an improvement for the standalone app, but I'm not sure yet if further changes are needed to solve our original issue on dotcom.

dmsnell:
why doesn't it work in the iframe? does `/me` return the wrong results?

kwight:
@dmsnell does 87963df work for you? (I ditched all the Prettier changes because madness – I should maybe do an update/once-over again in a separate PR.)

> why doesn't it work in the iframe? does /me return the wrong results?

I don't see the `/me` call firing (which would certainly explain it failing); will dig into it now with this new updated branch.

kwight:
Hm, I thought I had it with 7f4da13; this approach uses the same `wpcom` object with `wpcom-proxy-request` that the rest of the app is using. However, once built on dotcom, I'm not able to see the `/me` call now again, and the browser fails to load the iframe (but for no obvious reason).

kwight:
Like so many things, firing this up all fresh Monday morning, it now seems to be working perfectly.

standalone/auth-wrapper.jsx
```diff
@@ -85,10 +85,21 @@ export const AuthWrapper = Wrapped => class extends Component {
...
 
+    setTracksUser = () =>
+        this.state.wpcom
+        .me()
+        .get( { fields: 'ID,username' } )
+        .then( ( { ID, username } ) => {
+            window._tkq = window._tkq || [];
+            window._tkq.push( [ 'identifyUser', ID, username ] ) ;
+        } )
+        .catch( () => {} );
```
gwwar: on failure I think we should no-op any tracks requests, otherwise we'll still run into the current situation.

standalone/auth-wrapper.jsx
```diff
@@ -85,10 +85,21 @@ export const AuthWrapper = Wrapped => class extends Component {
...
+
     render() {
         const { clientId, redirectPath, ...childProps } = this.props;
 
         if (this.state.wpcom) {
+            this.setTracksUser();
```
gwwar: `componentDidMount`? Do we need to call this multiple times in render?

standalone/auth-wrapper.jsx
bperson: I'll add a "do not set cookie on `wp.com`" as a last mile safeguard in `w.js` which means you will get unreliable IDs across page reloads ( demultiplying `calypso_notifications_*` uniques ) if we continue to send pixels without an `identifyUser`, but we won't set cookies on `wp.com` no matter what. It's up to you folks to decide between: a bit more data ( raw counts ) with slightly unreliable uniques, and less data with 100% reliable uniques.

gwwar: Thanks @bperson we'll try and aim for more data with less reliable uniques. Let us know when the w.js update is ready and we can bump the cache break.

kwight: > on failure I think we should no-op any tracks requests, otherwise we'll still run into the current situation.

> It's up to you folks to decide between: a bit more data ( raw counts ) with slightly unreliable uniques, and less data with 100% reliable uniques.

I'd agree with @gwwar , if we don't have a valid authenticated user for some reason, we should just drop making the request. However, I don't see how we can tell if tracks considers the user known or not, _before_ making the `window._tkq.push()` call. We can't rely on the presence, absence, nor value of a `tk_ai` cookie, since it may not have been established yet (a successful `identifyUser` call just makes sure that we _don't_ have one). We could add something to the app's state upon the successful `me()` call (or use the presence of an existing `this.state.oAuthToken` to assume it's all good), but it would be nice to ask Tracks directly whether Tracks thinks the user is known or not.

@bperson could we either 1) have `identifyUser` be more helpful in how it handles cookies, or 2) add something to TKQ's API that we could access, eg. `window._tkq.userAuthenticated()`? Or maybe I'm missing an obvious easier solution?

bperson: @gwwar @kwight  If you're going for more data but less reliable uniques, you should just noop on error when fetching `/me`: Tracks will not use the logged-in user ID ( because ... we couldn't get one ;) ), create an anonymous ID at the first `recordEvent`, try to back it by a cookie which will fail because of the ( upcoming ) safeguard against `wp.com` in `w.js`. On subsequent reload ( or visiting any other site ), we won't be able to use the same anonymous ID because it wasn't persisted, so any new event will be associated with a completely different ID, leading to more unreliable uniques.

So if you're going with this solution, I think this PR looks good and there is nothing else to do on your end, except bumping the cache break in a couple of hours when I get the updated `w.js` out. 

Just for completing the thought process, if you were to go for less data with more reliable uniques, we would have to "block" sending `recordEvent` if we aren't able to `identifyUser`, the best way to accomplish that without encountering the state issue you mentioned @kwight ( not knowing if we've loaded an ID from `/me` when we push a `recordEvent` for instance ), is to let it push, but never load `w.js` to never send a pixel. And the easiest way to do that would be to load `w.js` on fetch success for `/me` instead of in the default template of the iframe content I believe.

kwight:
> Thanks @bperson we'll try and aim for more data with less reliable uniques. Let us know when the w.js update is ready and we can bump the cache break.

Ah, missed that, and yes, it sounds like if we wanted to go in the other direction at some point, it's enough work to warrant another PR anyways. I'll merge this as-is, and do bumps to the `w.js` version once that's happened..

PR #210, opened 2018-01-09
Remove `catch()` from the `/me` call to avoid nooping on all failures.

kwight:
@gwwar I believe we discussed this also around #209 – did I understand this correctly?


gwwar:
I think we can close this one out @kwight, generally we should be aware that a catch with catch any error from the chain not just the previous call (such as then)

PR #224, opened 2018-02-02
Give new post notifications the same top padding as comments and mentions.

kwight:
This PR adds the same top padding to new posts as we already give to comments and mention notifs. This can't be better generalized (eg. a selector of just `.wpnc__body .wpnc__body-content`), since it starts to mess up other types (such as likes). Note that this will also fix cross-posts (they have the same `.wpnc__new_post` class).

screen shot 2018-02-01 at 4 29 18 pm

_Before_

screen shot 2018-02-01 at 4 27 08 pm

_After_

Fixes https://github.com/Automattic/wp-calypso/issues/21558

**Testing**
 - Load this branch and verify there is a top padding on `.wpnc__body_content` for new post notifs, as well as a border between the content and the Like star.
 - Check that comments, x-posts, and auto-mattchers are treated the same.

### DIFF
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -744,16 +744,16 @@ body {
     padding: 0 $padding-large $padding-medium;
   }
 
-  .wpnc__comment .wpnc__body .wpnc__body-content {
-    box-shadow: inset 4px 0 0 lighten($gray, 30);
-    border-bottom: 1px solid lighten($gray, 30);
-    margin: 0 0 0 $padding-medium;
-    padding-top: $padding-small;
-  }
-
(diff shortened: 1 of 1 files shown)


dmsnell:
@kwight I think something happened here - I'm not sure that the large black boxes are ideal. Do you think we could maybe let the text shine through so we can read it?

kwight:
@dmsnell I was going for a minimal style, something simple and high contrast for a11y compatibility. Perhaps @drw158 could see about a version with softer grays?..

Copons:
I personally like the night theme, now I can finally read my notifications while in bed.

drw158:
I always make sure to use plenty of blackspace in my designs.

Issue #110, opened 2017-05-22
UI: Improve repeating filter loader 


roundhill:
If you switch to the `Unread` filter or other filters that don't have many notes, you'll see this:

![spinner-anim](https://cloud.githubusercontent.com/assets/789137/26318462/00a5c056-3ed0-11e7-9ff2-94827d02b25d.gif)

It'd be nice if it just animated normally, instead of resetting. Or for an easy fix we could just not show the spinner at all?

kwight:
The spinner could use some love; it's now visually clipped off on the sides too.

![screen shot 2018-01-11 at 12 47 49 pm 1](https://user-images.githubusercontent.com/349751/34846719-f6fbf866-f6cd-11e7-9b6b-b65534f4e13d.png)

PR #236, opened 2018-02-23
Replace the current spinner with Calypso's new pure CSS spinner.

### LINKED ISSUE #234: Spinner update from Calypso for consistency
In https://github.com/Automattic/wp-calypso/pull/22603 the spinner in Calypso was updated to have a more unique style but even more importantly to cut the CPU cost in half.

It would be relevant to update Notifications to use the same spinner (and maybe fix the "blinking" effect it does in the process if it's something simple to do).

cap-spinner-double

_(the spinner gif above glitches, see the [codepen](https://codepen.io/folletto/full/RQMEyd/) for a more accurate render)_

kwight:
This PR replaces the original poorly-performing spinner with a simpler CSS spinner, which was recently updated in Calypso.

spinner

Fixes #234 .

**Testing**
- Change both the `isLoading` prop [here](https://github.com/Automattic/notifications-panel/blob/master/src/templates/note-list.jsx#L371) and the `isSubmitting` prop [here](https://github.com/Automattic/notifications-panel/blob/master/src/templates/comment-reply-input.jsx#L49) to `true`, so you can see what's going on.
- Load notifications, and verify the note list spinner is correct.
- Reply to a comment, and verify the reply spinner looks correct.

### DIFF
--- a/src/boot/stylesheets/spinner.scss
+++ b/src/boot/stylesheets/spinner.scss
@@ -1,22 +1,18 @@
 .wpnc__main {
   .wpnc__spinner {
+    display: flex;
+    align-items: center;
     position: relative;
     top: 50%;
   }
 
-  // Start https://github.com/Automattic/wp-calypso/blob/master/client/components/spinner/style.scss
   @keyframes rotate-spinner {
     100% {
       transform: rotate( 360deg );
     }
   }
 
-  .spinner {
-    display: flex;
-    align-items: center;
-  }
-
-  .spinner__outer, .spinner__inner {
+  .wpnc__spinner__outer, .wpnc__spinner__inner {
     margin: auto;
     box-sizing: border-box;
     border: 0.1em solid transparent;
@@ -25,16 +21,15 @@
     animation-name: rotate-spinner;
   }
 
-  .spinner__outer {
+  .wpnc__spinner__outer {
     border-top-color: $blue-medium;
   }
 
-  .spinner__inner {
+  .wpnc__spinner__inner {
     width: 100%;
     height: 100%;
     border-top-color: $blue-medium;
     border-right-color: $blue-medium;
     opacity: 0.4;
   }
-  // End https://github.com/Automattic/wp-calypso/blob/master/client/components/spinner/style.scss
 }
--- a/src/boot/stylesheets/spinner.scss
+++ b/src/boot/stylesheets/spinner.scss
@@ -1,138 +1,40 @@
-//CSS Spinner http://codepen.io/MichaelArestad/pen/4f837dd51dc09256115afc9218edda2c
-@keyframes ui-wpnc__spinner-rotate-right {
-  0% {
-    transform: rotate(0deg);
+.wpnc__main {
+  .wpnc__spinner {
+    position: relative;
+    top: 50%;
   }
-  25% {
-    transform: rotate(180deg);
-  }
-  50% {
-    transform: rotate(180deg);
-  }
-  75% {
-    transform: rotate(360deg);
-  }
-  100% {
-    transform: rotate(360deg);
-  }
-}
 
-@keyframes ui-wpnc__spinner-rotate-left {
-  0% {
-    transform: rotate(0deg);
+  // Start https://github.com/Automattic/wp-calypso/blob/master/client/components/spinner/style.scss
+  @keyframes rotate-spinner {
+    100% {
+      transform: rotate( 360deg );
+    }
   }
-  25% {
-    transform: rotate(0deg);
-  }
-  50% {
-    transform: rotate(180deg);
-  }
-  75% {
-    transform: rotate(180deg);
-  }
-  100% {
-    transform: rotate(360deg);
-  }
-}
 
-$size: 20px;
-$transition-speed: .4s;
-.wpnc__spinner {
-  position: relative;
-  top: 50%;
-  left: 50%;
-  transform: translate(-50%, -50%);
-  border-radius: 100%;
-  width: $size;
-  height: $size;
-  background: lighten($gray, 20);
+  .spinner {
+    display: flex;
+    align-items: center;
+  }
 
-  // before & after make the donut hole
-  &:before, &:after {
-    content: '';
-    position: absolute;
-    background: $white;
+  .spinner__outer, .spinner__inner {
+    margin: auto;
+    box-sizing: border-box;
+    border: 0.1em solid transparent;
     border-radius: 50%;
+    animation: 3s linear infinite;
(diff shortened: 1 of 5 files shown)


kwight:
screen shot 2018-02-23 at 4 00 20 pm
screen shot 2018-02-23 at 3 57 25 pm

kwight:
This also tidies up https://github.com/Automattic/notifications-panel/issues/110#issuecomment-357057062.

kwight:
@folletto sorry, just the cropping; the resetting I imagine to be an issue around constant re-renders, which I'll need to dig into separately.

folletto:
Understood! Thank you! :)

src/boot/stylesheets/spinner.scss
```diff
@@ -1,138 +1,40 @@
...
-    left: 0;
 
-    &:before {
-      left: 100%;
-      border-top-left-radius: 0;
-      border-bottom-left-radius: 0;
-      transform-origin: 0 50%;
-      transition-delay: 0;
-    }
+  .spinner {
```
gwwar: Out of curiosity is there a reason why we don't namespace all of our classes? Or did we feel the parent class was enough?

src/templates/comment-reply-input.jsx
```diff
@@ -286,10 +286,7 @@ const CommentReplyInput = createReactClass({
 
     if (this.state.isSubmitting) {
       submitLink = (
-        <div className="wpnc__spinner animated">
-          <span className="side left" />
-          <span className="side right" />
-        </div>
+        <Spinner className="wpnc__spinner" />
```
gwwar: Let's have `wpnc__spinner` be a default class

src/boot/stylesheets/spinner.scss
dmsnell: when it was young we were in our own iframe and the app was small and we had no felt need to namespace.

gwwar: @dmsnell would we like to namespace the rest? I feel like that's usually the safest

gwwar:
Looks like the parent is being rendered multiple times here which is resetting the animation. Note that setting the following doesn't work https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode

rerender

gwwar:
🤷‍♀️ Did I never notice before that our polling behavior grabs 10, then 20, ... up to 100 notes for polling?

screen shot 2018-02-27 at 2 50 51 pm

As far as I can tell it's starting from the same note range. Eg note 0 is the same for the first request as it is for the 10th one.

src/boot/stylesheets/spinner.scss
dmsnell: yeah that's great. we started using `wpnc_` as the prefix. no reason not to add it everywhere else. I predict that we only put it where we needed to in order to integrate this code directly into Calypso

dmsnell:
> Did I never notice before that our polling behavior grabs 10, then 20, ... up to 100 notes for polling?

This is the last major bit of unfinished business I have on my todo list for the notes client. It's not trivial to adjust. First of all, the reason we always grab the newest notes is because they might be changing and we wouldn't want to miss a new update because we're grabbing the old ones.

The thing to realize is that the app state is a nightmare. I don't think it would take another full week's meetup to change but maybe a week of one person dedicated to the cause.

kwight:
For the rendering issue, let's tackle it in a separate PR addressing #110 .

> Out of curiosity is there a reason why we don't namespace all of our classes? Or did we feel the parent class was enough?

Since we're using Calypso's spinner, my thought was towards keeping the code the same between the two repos (hence the `// start` and `// end` comments); otherwise, any changes to Calypso's styles and markup mean updating the style prefix every time when porting those changes to this repo (although, I'm the first to admit this isn't a lot of unreasonable work). Strong feelings either way?

> Let's have wpnc__spinner be a default class

Not sure what you mean here?

vindl:
>>Let's have wpnc__spinner be a default class

>Not sure what you mean here?

My guess would be to move it to component render method instead of passing it as a prop.
https://github.com/Automattic/notifications-panel/pull/236/files#diff-a555cad79b68b4f00c9b34c45ac49947R359

kwight:
OK, I prefixed classes and set `wpnc__spinner` as the default class, we should be ready to go.

src/templates/comment-reply-input.jsx
vindl: Now that it's set as default, we no longer need to pass the class here?

vindl:
> Since we're using Calypso's spinner, my thought was towards keeping the code the same between the two repos (hence the // start and // end comments); otherwise, any changes to Calypso's styles and markup mean updating the style prefix every time when porting those changes to this repo (although, I'm the first to admit this isn't a lot of unreasonable work). Strong feelings either way?

Since we are using Sass could we extend from existing class, so we can have it namespaced and avoid duplication at the same time? (I'm guessing that the answer is no because of separate codebases, and all the different places that notes are used in)

```scss
.wpnc__spinner {
    @extend .spinner;
    ...
}
```

If not, then I think that adding a small maintenance burden by having everything prefixed consistently is fine. Should we also add the comments in Calypso counterpart in order to inform future devs to contact us for syncing if they decide to change the relevant parts of the spinner?

folletto:
> I'm guessing that the answer is no because of separate codebases, and all the different places that notes are used in

That would be the reason for my "no" too. I feel it's better to have a risk to have a component outdated than the risk of breaking it entirely for a missing reference that isn't automatically enforced.

> Should we also add the comments in Calypso counterpart in order to inform future devs to contact us for syncing if they decide to change the relevant parts of the spinner?

This sounds like a good idea! Like "Open a PR also on `notifications-panel` if the component gets updated" kind of thing?

src/templates/comment-reply-input.jsx
gwwar: Yup we can remove this

gwwar:
> Since we are using Sass could we extend from existing class

Let's avoid `extend`, overall I find it hard to maintain (media queries don't work) and it's super easy to add css selector bloat very easily. If we want to use the spinner as a generic component, we can pull it out of Calypso entirely and use it as an npm dependency.

Let's 🚢 this for now, and see if we need to pull out that component as a dependency later. I don't think the spinner will be worked on that often, so we may not need to do htis.

vindl:
> This sounds like a good idea! Like "Open a PR also on notifications-panel if the component gets updated" kind of thing?

Yes, that's exactly what I had in mind. :+1:

kwight:
Instructions added to Calypso in https://github.com/Automattic/wp-calypso/pull/23145.

PR #239, opened 2018-03-13
Readme: adds tag and release documentation

rodrigoi:
Adds tag and release instructions to the `README.md`.


README.md
```diff
@@ -48,3 +48,28 @@ npm start
 After it boots up load the entry at [notifications.localhost:8888](notifications.localhost:8888) in your browser.
 
 [react]: https://facebook.github.io/react/
+
+### Tag, Release and Publish
+
+To create a new version, you can run this command
+
+```bash
+npm version npm version [<newversion> | major | minor | patch] --no-git-tag-version
```
Copons: This could probably lose an `npm version`. 🙂

Copons:
This looks good, even though it doesn't have enough Star Wars jokes inside.

README.md
```diff
@@ -48,3 +48,28 @@ npm start
...
+```
+
+that will update the `package.json` and the `package-lock.json` files with the new version. The `npm version` command can also create a new commit and a tag for us, but it does not follow our naming conventions.
+Then, you need to commit the changes and create a new git tag. Replace `<newversion>` with the version generated by `npm version`.
+
+```bash
+git commit -a -m "Bump to version <newversion>"
```
gwwar: Probably missing a step here, to review/merge the PR.

Similarly we'd probably want to tag from master vs the branch.

rodrigoi:
Closing this because we already have a _Field Guide_ document.

gwwar:
Hmm, doesn't hurt to point at documentation in the README. What if we copied relevant copy from the post and added gh shorthand to link to it? It may be slightly more of a pain to update documentation, but I don't see this changing that often. What do folks think @rodrigoi @kwight ?

kwight:
I see it as duplicating data into a public repo that's not relevant to everyone, but I'm happy to disagree and commit 😊

gwwar:
We could at a minimum point to the fg page in gh shorthand. At least for me it's not the first place to look for instructions.

PR #246, opened 2018-04-04
Add/offline gridicon

enejb:
Currently we don't have a offline gridicon. 
The icon is not something that is pressent as a noteicon. So in order to get the icon to appear we will have to set the new icon slug in the api. 

In https://github.com/Automattic/jetpack/issues/7892#issuecomment-375001662 it was suggested that we used the offline gridicon instead of the current (default) icon for the notification. 

This PR one of the steps in adding the icon to notification. 

See: screen shot 2018-04-04 at 3 29 40 pm

This PR does not fit the whole issue just adds the missing icon

### DIFF
--- a/src/utils/noticon2gridicon.js
+++ b/src/utils/noticon2gridicon.js
@@ -16,6 +16,7 @@ export const noticon2gridicon = c =>
       '\uf414': 'warning',
       '\uf418': 'checkmark',
       '\uf447': 'cart',
+      'offline': 'offline'
     },
     c,
     'info'
--- a/src/boot/stylesheets/main.scss
+++ b/src/boot/stylesheets/main.scss
@@ -121,6 +121,10 @@ body {
   .gridicon {
     fill: currentColor;
   }
+  
+  .gridicons-offline {
+    margin:1px 0 0 1px;
+  }
 
   .wpnc__user__username, span.wpnc__user {
     font-weight: 600;
--- a/src/templates/gridicons.jsx
+++ b/src/templates/gridicons.jsx
@@ -36,6 +36,14 @@ export default class extends React.Component {
           </svg>
         );
 
+      case 'gridicons-offline':
+        return (
+          <svg {...sharedProps} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+            <rect x="0" fill="none" width="24" height="24"/>
+            <g><path d="M10 3h8l-4 6h4L6 21l4-9H6l4-9"/></g>
+          </svg>
+        );
+
       case 'gridicons-spam':
         return (
           <svg {...sharedProps} xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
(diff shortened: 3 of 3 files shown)


kwight:
What is this for? How do we test it? Any more details?

enejb:
I updated the PR's description to add more info and context.

kwight:
I ended up just adding a stub that means the noticon2gridicon translation just slides through (it also requires D11544-code). It'll need more thinking moving forward in a Noticon-free world, but this gets us going for now.

kwight:
Paired with d11544-code, we get this. Maybe you want to center it a little better too?

screen shot 2018-04-05 at 2 16 01 pm

enejb:
I think this is ready to merge :shipit:

src/utils/noticon2gridicon.js
```diff
@@ -16,6 +16,7 @@ export const noticon2gridicon = c =>
       '\uf414': 'warning',
       '\uf418': 'checkmark',
       '\uf447': 'cart',
+      'offline': 'offline'
```
dmsnell: this is a bit unsettling. did we check with the mobile teams before sending notifications with a non-noticon point reference?

kwight: Indeed, this was one of today's tasks :) It looks like the mobile apps (at least iOS which I checked) uses a Noticon font, so indeed this may be problematic; posted for some guidance in p77Llu-a5H-p2.

kwight: I hardcoded the offline icon in XCode today to see what would happen, and it was basically what we expected: a broken icon, because it's not a recognizable character in the Noticon font:

screen shot 2018-04-12 at 11 14 21 am

kwight:
So, we have a few options here.

1. Don't use the `offline` icon, but stick with something in the Noticon font. @rickybanister if there's any urgency at all, this may be the way to go for now 🙃
2. Introduce a fallback. From what I can tell, this would be not be trivial in the apps themselves, but could be possible on the API side if we can identify the calling client as an app vs web-based. This could be done in `class.noticon-utf8-values.php` server-side in a new `get_offline()` method (from d11544-code).
3. Have a separate mechanism for Gridicon icons; rather than having just `note.noticon`, we add `note.gridicon`, which apps handle with a fallback Noticon icon.

None of these sound great to me :) Anyone have any opinions, @dmsnell @drw158 ?

rickybanister:
Number 1 sounds the 'easiest' for now, but I don't know what icon would be better, perhaps just the `(!)` since monitor is warning you. 

Number 3 sounds like the most future-proof way to go as it would help us to slowly transition away from noticons over time. 

It's all y'all's call.

kwight:
@rickybanister cool, agreed, #3 makes sense as a transition path which I can start work on. I'll keep this open to keep working on, but if any urgent changes are needed, good to fall back to the `(!)` icon then.

kwight:
Adding Gridicons that don't exist in Noticons will require adding a new `gridicon` property to notes, and then patching the mobile apps to look for it (and falling back to the `noticon` handling I guess?). The mobile apps also continue to use a Noticon font, so that needs to be dealt with on their end at some point.

There's no urgency for any of this, so I'll close and it can be picked up later if found important enough at some point.

PR #253, opened 2018-04-04
Improve link destinations for empty filters.

kwight:
Suggested by @designsimply and @rachelmcr in https://github.com/Automattic/notifications-client/issues/609.

This updates the empty Likes text to link to https://wordpress.com/activities/likes , and the empty Comments to https://wordpress.com/read/search

**Testing**
* Start notifications from this repo.
* Log in as a user with no Likes or Comments.
* Verify the content goes to its proper destination.

### DIFF
--- a/src/templates/filters.js
+++ b/src/templates/filters.js
@@ -93,7 +93,7 @@ export const Filters = {
       emptyLinkMessage: i18n.translate("Get noticed: comment on posts you've read.", {
         comment: 'Notifications likes filter: no Notifications',
       }),
-      emptyLink: 'https://wordpress.com/posts/',
+      emptyLink: 'https://wordpress.com/activities/likes/',
 
       filter: ({ type }) => 'comment_like' === type || 'like' === type,
     };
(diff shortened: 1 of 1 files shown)


src/templates/filters.js
```diff
@@ -75,7 +75,7 @@ export const Filters = {
       emptyLinkMessage: i18n.translate("Get noticed: comment on posts you've read.", {
         comment: 'Notifications follows filter: no notifications',
       }),
-      emptyLink: 'https://wordpress.com/read/',
+      emptyLink: 'https://wordpress.com/activities/likes/',
```
Copons: The PR description says this should be for the `likes` section, but it's in `follows` instead.

Though, both sections use the same empty message, and they should have the same link as well.

Copons:
Minor nitpick quickly fixable that I accidentally submitted as change request. Otherwise this LGTM!

src/templates/filters.js
kwight: Oh! Well spotted, and good suggestion; added the same to Likes as intended, and left this one here for Follows 👍

PR #276, opened 2018-05-25
Packages: Update Hoek dependencies.

kwight:
This PR updates the repo to Webpack 3. `webpack-dev-server` also gets bumped to 2.11.2.

**Testing**

1. Blow away `node-modules` and do a fresh `npm install`.
2. Smoke test as much of the app as possible; there should be no related errors or regressions.
3. There should be no references to the `hoek` package in `package-lock.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,8 +66,8 @@
     "redux": "3.6.0",
     "sass-loader": "6.0.3",
     "style-loader": "0.16.0",
-    "webpack": "2.3.1",
-    "webpack-dev-server": "2.4.2",
+    "webpack": "^3.12.0",
+    "webpack-dev-server": "^2.11.2",
     "wpcom": "5.3.0",
     "wpcom-proxy-request": "3.1.0"
   }
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,12 +17,12 @@
       "dev": true
     },
     "accepts": {
-      "version": "1.3.4",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -70,5 +70,11 @@
     "webpack-dev-server": "2.4.2",
     "wpcom": "5.3.0",
     "wpcom-proxy-request": "3.1.0"
+  },
+  "dependencies": {
+    "boom": "^7.2.0",
+    "cryptiles": "^4.1.1",
+    "fsevents": "^1.2.4",
+    "hawk": "^7.0.7"
```
dmsnell: these probably belong above in `devDependencies` no?

also, why are we _adding_ these? I think we want to keep dependencies-of-dependencies out of `package.json` and update instead the primary dependencies which are pulling in the vulnerable version of `hoek`

kwight: Wuuuuuuttt??!! Haha, I never even noticed I was adding these to `package.json` – I verified what I wanted to see in the package lock file and pushed 🙃Thanks, yes, I agree it should be just updating the dependencies themselves, will sort it out on Monday.

kwight:
@dmsnell I think this should take care of it now.

kwight:
> if you regenerated the package-lock.json file from scratch then :shipit:

Yep; tested by confirming the changes worked fine, then blew away both `node_modules` and `package-lock.json`, then did a fresh install and verified again. Thanks for the review!

PR #280, opened 2018-06-12
Update engines to node 10.4 and update dev dependencies using npm audit fix

gwwar:
This PR updates our expected major version of Node to help with ease of development, and also updates a few stale npm developer dependencies using `npm audit fix`.

### Testing Instructions

Test Development Container:
- `rm -rf node_modules`
- switch to Node v10.4
- `npm install`
- `npm start`
- Visit http://notifications.localhost:8888/
- Verify that there are no regressions.

Calypso:
- navigate to your wp-calypso checkout
- checkout `update/notifications`
- `npm start`

or use the live link in 
https://github.com/Automattic/wp-calypso/pull/25466

- Verify that there are no regressions.

Standalone (iframe):
- sandbox a site and widgets (dm if you need more specific instructions)
- run `~/public_html/bin/notes/build-notifications.sh -c2.2.0-alpha.1`
- visit a blog site, and open the masterbar notifications there (not Calypso)
- verify that there are no regressions

see also PCYsg-elI-p2

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "2.2.0",
+  "version": "2.2.0-alpha.1",
   "description": "The core notifications panel for WordPress.com notifications",
   "main": "src/Notifications.jsx",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "notifications-panel",
-  "version": "2.1.10",
+  "version": "2.2.0",
   "description": "The core notifications panel for WordPress.com notifications",
   "main": "src/Notifications.jsx",
   "scripts": {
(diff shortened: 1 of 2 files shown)


gwwar:
I did some light smoke testing in all three envs, I'll abandon Automattic/wp-calypso#25466 if this lands. 👀 appreciated!

package.json
```diff
@@ -18,8 +18,8 @@
     "Opera 12.1"
   ],
   "engines": {
-    "node": ">= 8.9.1",
-    "npm": ">= 5.5.1"
+    "node": ">= 10.4.0",
```
dmsnell: well now I guess we should do `10.4.1` after the security update?

gwwar: Yup, I will. The >= will accept that as well, so we probably only need to update engines again after the next Major version bump.

gwwar: I'll keep it here since https://github.com/Automattic/wp-calypso/pull/25469 hasn't landed yet. but I don't think this will be an issue.

PR #281, opened 2018-06-13
Add Jest and Remove Mocha

gwwar:
This PR removes an old `mocha` version and switches out test runner for `jest`. By removing `mocha`, we also get rid of a nagging dev dependency warning with it's old sub-depency, `growl`. I also bumped `debug` to remove a number of npm audit warnings.

We had one testfile, but since we did not setup CI, I feel like we can disable this for now and re-enable the test and needed webpack/babel tinkering in a future PR. After we re-enable real tests, let's also hook this up to Circle CI

### Testing Instructions
- `npm start`
- No regression in the local testing env http://notifications.localhost:8888/
- Paste `localStorage.setItem('debug', '*')` in the console of your local test env. debug statements still work
- `npm test` runs, but tells us we have no tests!


codebykat:
This builds and tests out fine (with the expected "no tests found" complaining from jest).

`npm audit` is still complaining about "13 vulnerabilities (4 low, 9 moderate)".

gwwar:
Thanks for the review @codebykat !

PR #282, opened 2018-06-14
[WIP] Dependencies: Upgrade i18n-calypso and wpcom-proxy-request

codebykat:
as per

```                       === npm audit security report ===

┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ debug                                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ i18n-calypso [dev]                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ i18n-calypso > debug                                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/534
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1878,6 +1878,12 @@
         }
       }
     },
+    "charenc": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/charenc/-/charenc-0.0.2.tgz",
+      "integrity": "sha1-wKHS86cJLgN3S/qD8UwPxXkKhmc=",
+      "dev": true
+    },
     "chokidar": {
       "version": "2.0.3",
       "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-2.0.3.tgz",
@@ -2382,6 +2388,12 @@
         "shebang-command": "^1.2.0",
         "which": "^1.2.9"
       }
+    },
+    "crypt": {
+      "version": "0.0.2",
+      "resolved": "https://registry.npmjs.org/crypt/-/crypt-0.0.2.tgz",
+      "integrity": "sha1-iNf/fsDfuG9xPch7u0LQRNPmxBs=",
+      "dev": true
     },
     "cryptiles": {
       "version": "2.0.5",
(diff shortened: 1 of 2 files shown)


gwwar:
🤔 Calypso is also on 1.8.4, which doesn't have this issue. I wonder if it's because we're missing the "dependencies" prop on package.json that we're seeing odd status messages. Otherwise we could target the moderate warnings first.

codebykat:
Investigation results for the remaining issues:

Of the 13 vulnerabilities reported by `npm audit`, 9 are related to node-sass:

* 8/13 are hoek via node-sass. Open issue here: https://github.com/sass/node-sass/issues/2355
* 1/13: tunnel-agent via node-sass. "This dependency is only used for our install and not for any runtime code" as per issue here: https://github.com/sass/node-sass/issues/2164

The remaining 4 are dependencies of our other packages:

* 3/13: debug via i18n-calypso. rRecommended: i18n-calypso should upgrade debug to >= 2.6.9 < 3.0.0 || >= 3.1.0
* 1/13: superagent via wpcom > wpcom-xhr-request. Recommended: upgrade superagent to >=3.7.0.

codebykat:
>  I wonder if it's because we're missing the "dependencies" prop on package.json that we're seeing odd status messages from github otherwise we could target the moderate warnings first.

Hmmmm so when I upgraded `wpcom-proxy-request` it told me to also upgrade `webpack` and `ajv`; somewhere in there it did add the dependencies prop, but with only `ajv` in it.

Should we pull in all the existing dependencies from package.lock? I glanced at the prop in Calypso and there's quite a lot in there. Is there a way to generate it or something?

codebykat:
@dmsnell if you have a second maybe you can re-review / answer the above? :)

dmsnell:
> Should we pull in all the existing dependencies from package.lock? I glanced at the prop in Calypso and there's quite a lot in there. Is there a way to generate it or something?

@codebykat we should be able to delete that file and the `node_modules` directory and then run `npm install` - it should regenerate it and that's a common way to resolve issues like this.

dmsnell:
on the further point I think we should be fine doing whatever updates we need if they don't involve too much work. depending on what `webpack` update it wants that could mean easy or hard.

within the next two weeks I'd like to see if I can't merge this codebase into Calypso and then in that case we'd end up blowing away all this duplicate maintenance. point being: don't spend too much time on updating packages or config files; if it's an easy quick win then 💯 go for it, but if not then I'll see if we can't just eliminate the effort entirely 👍

codebykat:
Updated with a new package-lock.json. Guess I'll merge and see what happens :)

PR #283, opened 2018-06-22
Fix assignment to undeclared variable in state

blowery:


src/state/index.js
```diff
@@ -20,7 +20,7 @@ export const init = ({ customEnhancer, customMiddleware = {} } = {}) => {
   const middle = withMiddleware(customMiddleware);
   const create = customEnhancer ? customEnhancer(middle) : middle;
 
-  store = create(reducer, reducer(undefined, { type: '@@INIT' }));
+  const store = create(reducer, reducer(undefined, { type: '@@INIT' }));
```
dmsnell: brought in in https://github.com/Automattic/notifications-panel/pull/108/files#diff-1dabc0c5cbcbadc986b3d7e221b63734

maybe it would be more in the spirit of the original code change to just return `create( … )`. I don't see any need for the `store` to have its own binding.

PR #288, opened 2018-08-17
Start updating repository to move towards closing it down.

dmsnell:
In this PR I'm trying to quickly prevent people from doing work here while we merge this code into Automattic/wp-calypso

 - remove the build scripts
 - add a note in the README

The point is to prevent someone from contributing here after https://github.com/Automattic/wp-calypso/pull/26757 merges as we want them to contribute to Calypso directly


vindl (approved):
LGTM! :shipit:

Maybe you could add a pre-commit check that prevents committing with appropriate message, but that's completely optional.

